PR #35, opened 2020-06-14
fix conversation result usage in loadEntireConversation

banilaste:
In `loadEntireConversation`, after loading a conversation with `loadConversation`, the resulting `json` variable was an array of `[expectedJson, request]` (from `q.callback([json, q.queryJSON])` in WhatsAppWeb.Recv.js) and was expected to be `expectedJson` only by the following code.


adiwajshing:
Makes sense. That you for this bug fix!

PR #52, opened 2020-07-03
Implements QR Code refresh

kaykyr:
Implemented a function to refresh QR Code automatically after 20 seconds and stop refreshing after 5 tries, like in WhatsApp Web.


adiwajshing:
Looks good. Can you revert your commits on `example.ts` & ignore `yarn.lock`?

kaykyr:
> Looks good. Can you revert your commits on `example.ts` & ignore `yarn.lock`?

Sure!

PR #53, opened 2020-07-04
Implements QR Code refresh

kaykyr:


src/WAConnection/Validation.ts
```diff
@@ -73,6 +74,15 @@ export default class WAConnectionValidator extends WAConnectionBase {
                 return this.userMetaData
             })
     }
+
+    /** Refresh QR Code */
+    protected refreshQrCode() {
+        const data = ['admin', 'Conn', 'reref']
+        return this.query(data).then((json) => {
```
adiwajshing, line 81: Put this in a single line?

src/WAConnection/Validation.ts
```diff
@@ -152,13 +162,24 @@ export default class WAConnectionValidator extends WAConnectionBase {
...
             this.authInfo.clientID,
         ])
-        return this.waitForMessage('s1', [])
+
+        try {
+            const response = await this.waitForMessage('s1', [], 20 * 1000)
+            return response
+        } catch (err) {
+            if (this.loops < 5) {
+                this.loops++
```
adiwajshing, line 178: Any particular reason you made `loops` a class variable and not a local functional variable?

src/WAConnection/Validation.ts
```diff
@@ -152,13 +162,24 @@ export default class WAConnectionValidator extends WAConnectionBase {
...
         this.curveKeys = Curve.generateKeyPair(Utils.randomBytes(32))
         this.onReadyForPhoneAuthentication([
             ref,
             Buffer.from(this.curveKeys.public).toString('base64'),
             this.authInfo.clientID,
         ])
-        return this.waitForMessage('s1', [])
+
+        try {
+            const response = await this.waitForMessage('s1', [], 20 * 1000)
```
adiwajshing, line 174: Also, I'm fairly certain the message tag (`s1`) will change after a re-ref, so waiting for that message will never end. 

Does the tag not change?

kaykyr: This is to wait for QR Code scanning confirmation, so while confirmation is not ready the script will keep refreshing QR Code until this message response, or until max loops reached.

src/WAConnection/Validation.ts
kaykyr: Sorry, my bad. I'll resend pull request without as class variable.

src/WAConnection/Validation.ts
kaykyr: It's better to keep clean code.

PR #54, opened 2020-07-04
Implements QR Code refresh + cmd listener

kaykyr:
I removed loops from the class, implemented in the same method.
Please, test it, and you will know that the implementation of `reref` is correct.

Please, feel free to change code as you prefer, this is a very important feature and I hope you can merge this pull request.


adiwajshing:
That's a cool idea -- listening for a `cmd` & renewing the session based on that. However, I believe it can be better implemented. If you could implement it in the following way, that would be great:

Add a function to the class: `setOnLoggedOut (callback: () => boolean)`, so when the client gets disconnected, the callback is called & the **client is closed**

When a user implements the callback using:
``` ts
    client.setOnLoggedOut (() => {
        console.log('oh no I got logged out!')
        return true
   })
```
if callback returned true, the client should begin the reconnecting process again.

This would make the functionality more extensible & user friendly.

kaykyr:
> That's a cool idea -- listening for a `cmd` & renewing the session based on that. However, I believe it can be better implemented. If you could implement it in the following way, that would be great:
> 
> Add a function to the class: `setOnLoggedOut (callback: () => boolean)`, so when the client gets disconnected, the callback is called & the **client is closed**
> 
> When a user implements the callback using:
> 
> ```ts
>     client.setOnLoggedOut (() => {
>         console.log('oh no I got logged out!')
>         return true
>    })
> ```
> 
> if callback returned true, the client should begin the reconnecting process again.
> 
> This would make the functionality more extensible & user friendly.

Hello! Sorry latency! I'll update this pull request asap!

PR #59, opened 2020-07-09
Update Utils.ts

kaykyr:
Change the way we create message tags to index model. Index model is more reliable than timestamp model, because we can request many times in the same second. Sending requests with the same message tag causes freezes in the script; WhatsApp ignore many massages with the same message tag, creating a index we'll always send a unique message tag.


adiwajshing:
Cool fix. However, since the 'index' is already tracked in `WAConnection` as the `msgCount` property, can you implement it like this:

modify `generateMessageTag` to include an `epoch` argument
``` ts
export function generateMessageTag(epoch?: number) {
    let tag = new Date().getTime().toString() 
    if (epoch) tag += '-' + epoch // attach epoch if provided
    return tag
}
```

Then, in `WAConnection`, pass `this.msgCount` to the `generateMessageTag` function in the `sendBinary` & `sendJSON` functions.

Would be cleaner & more in line with the compartmentalized design of Baileys.

Thanks.

PR #64, opened 2020-07-12
Log correct err object

SvanBoxel:
`error` isn't defined in the catch callback, but `err` is.  Changed it to the correct name this in console logging method.


adiwajshing:
So, `ConversationExtract.js` hasn't been updated for Baileys 2.0 which makes this fix kinda redundant because the file itself won't run. However, have you managed to make it work?

adiwajshing:
Closing for no activity

PR #81, opened 2020-07-26
add mentionedIds to messages

gbrune:
This PR adds the possibility to include mentioned user ids in the message (works with text messages and image captions).
Phone numbers of mentioned users are replaced with their contact name (if in contacts)
Test is included.

**Example:**

Both messages have exactly the same message text, the first one has an additional `options.mentionedIds` field.
whatsapp-mentions

**Usage:**

```TypeScript
const options: MessageOptions = {
mentionedIds: ["123@s.whatsapp.net", "456@s.whatsapp.net"]
}
const messageText = "hello @123, how is @456?" // only use @ + phonenumber in message, no @s.whatsapp.net!
await this.client.sendMessage("123@s.whatsapp.net", messageText, MessageType.text, options)
```


adiwajshing:
So, this can already be done in the library:

``` ts
const message: WATextMessage = {
    text: 'hello @123 and @345', 
    contextInfo: { 
        mentionedJid: ['123@c.us', '345@c.us'] 
    }
}
client.sendMessage ('1234', message, MessageType.text)
```

gbrune:
Oh wow, I totally missed that. :/
Does this work with mentions in image captions too?

adiwajshing:
You could but it's a more roundabout way to go about it.

So, in the latest commit, you can:
``` ts
const options = {
    caption: 'hello @234',
    contextInfo: { mentionedJid: [ '234@s.whatsapp.net' ] }
}
client.sendMessage ('1234', buffer, MessageType.image, options)
```

gbrune:
Thanks!
Closing this PR.

PR #104, opened 2020-08-09
Added typedoc. Runnable with yarn build:docs

AZMCode:
Signed-off-by: AZMCode <adrianozambrana@protonmail.com>


AZMCode:
There still is the question of how these can/should be accessed.

Issue #111, opened 2020-08-17
[Feature suggestion] Package and release with CI


AZMCode:
As it stands right now, the npm package builds itself on the target machine, instead of before publishing. Github actions could automatically build and package the specific files needed for a significant boost with respect to install time.

Github Actions could also build new documentation every time a push is made to master, and publish the docs in the project's Github Pages site.

I'm aware this is low priority, just thought to lease this here for when there's enough time.

adiwajshing:
I really love the idea. Right now, I'm planning a V3 release with a bunch of really nice features like automatic chat & contact maintainence, phone pinging & an event emitter based syntax.

It would be great if you could contribute to the GitHub actions!

AZMCode:
Ill do my best in my free time, chief!
I actually think i have a template for this

PR #112, opened 2020-08-17
GitHub Actions integration

AZMCode:
The repo will now automatically put the typedoc results in the `gh-pages` branch of the repo.
A new push to master will result in a new Release with a release asset to match and, if the `NPM_TOKEN` secret is set will also upload the package to npm. Fulfills #111 

Note: Could not test the jobs that relied on a push to master, for obvious reasons.


adiwajshing:
Looks really good, thanks a lot for this! If I set the `NPM` token, it'll publish to NPM on every commit?

AZMCode:
It should publish to NPM on push to master. The secret is `NPM_TOKEN`
Please consider this was done rather quickly, and the push to master might not work just yet. I'm trying to find a way to test it. Otherwise, the `Test` job works.

AZMCode:
Nope, automated testing is broken. Since the program requires logging into Whatsapp to test, we can't run it automatically
We could store the session keys in the secrets, but then that'd mean having a dedicated phone just for CI

AZMCode:
Forgot to mention, this PR creates a new release and uploads a package to it every time there's a push to master.

AZMCode:
Nope, PR is broken rn. Just tested it and it didnt run the publishing jobs

AZMCode:
Alright, I've tested it a bit more on my fork, and it all seems to work.

adiwajshing:
So, I'm going to include this for Baileys 3.0, is there a way to not release with every push?

AZMCode:
I believe it is possible to remove the "push" trigger, and then manually execute the workflow when needed.
Might give it a go.

AZMCode:
Done. Workflow will now only work when manually activated. Under `Actions`->`CI`->`Run Workflow`

adiwajshing:
Merged, thanks a lot @AZMCode

adiwajshing:
@AZMCode so, the publish package is still called on every push causing the workflow to be marked as failed. Can you see what's up?

AZMCode:
Oh lord. it doesn't name the release correctly.
Debugging on master sounds like a not fun time.

AZMCode:
Already made a new PR which hopefully should fix all these issues.

PR #139, opened 2020-09-07
fix: call loadMessages multiple times

galiazzi:
When call loadMessages multiple times the count its not respected. Example:

```js
const { messages } = await this.client.loadMessages("jid", 5);
// messages contains last 5 messages

// calling again with a smaller count
const { message } = await this.client.loadMessages("jid", 2);
// 5 messages are returned instead of 2
```


adiwajshing:
Cool, thanks for this.

PR #203, opened 2020-10-25
Fix calculating audio duration.

### LINKED ISSUE #189: [BUG] Audio duration is 0:00
**Describe the bug**
The audio is being sent successfully. Anyway, it doesn't seems to have the duration label in a conversation.

**To Reproduce**
1. Created a new connection
2. Send an audio message using Baileys

**Expected behavior**
The user should receive the audio message, and it also needs to have the duration from the audio.

**Environment (please complete the following information):**
 - Is this on a server? Yes
 - Do you have multiple clients on the same IP? Yes
 - Are you using a proxy? No

**Additional context**

We're trying to send an audio message using this API. The audio is sent successfully, anyway, it does not seems to have the duration label attached to the audio object. 

**Reproducible example**

Send an audio with the following audio URL: https://file-examples-com.github.io/uploads/2017/11/file_example_MP3_700KB.mp3

The code i'm using is the following:

```
const id = (to.indexOf('-') > -1 ? `${to}@g.us` : `${to}@s.whatsapp.net`) // the WhatsApp ID
let buffer = fs.readFileSync('test.mp3'); // This is the file I downloaded from https://file-examples-com.github.io/uploads/2017/11/file_example_MP3_700KB.mp3

const message = await conn.sendMessage(id, buffer, MessageType.audio, {
  mimetype: Mimetype.ogg
});
```

Also, tested with **MessageType.audio** as messageType and **mimetype = Mimetype.mp4Audio**, with same results.

### Image capture

IMG_4658

Notice that the message is sucessfully sent, and it also is playable. It seems to work everything right. But the duration is 00:00 even when the user is playing the audio.

What am I doing wrong? Any ideas?

Thanks.

Borewit:
Replace get-audio-duration & mp3-duration with [music-metadata](https://github.com/Borewit/music-metadata).

Pure JavaScript solution, eliminates dependencies on native utilities like [ffprobe](https://ffmpeg.org/ffprobe.html).

Please note:
- I have not tested the application with this PR. But I know [music-metadata](https://github.com/Borewit/music-metadata) will give you the right duration.
- Pass the right mime-type to `export async function getAudioDuration (buffer: Buffer) {`, and passing that `musicMetadata.parseBuffer` would improve performance (skip automatic content detection). I you like I can add that,
- The code us full of unnecessary trailing spaces, this makes it hard to commit clean PRs. 

Fixes #189.


adiwajshing:
@Borewit your library looks really good & I would love to use it. However, I would greatly appreciate it if you could test the code out before I merge

adiwajshing:
Okay the tests are passing. Merging. Thanks

PR #208, opened 2020-10-28
Audio decoding based on MIME-type.

Borewit:
Use the MIME-type to determine the audio type.
Without the MIME-type [music-metadata](https://github.com/Borewit/music-metadata) will use [file-type](https://github.com/sindresorhus/file-type) to determine the audio type.

Related to #203

### DIFF
--- a/src/WAConnection/6.MessagesSend.ts
+++ b/src/WAConnection/6.MessagesSend.ts
@@ -120,7 +120,7 @@ export class WAConnection extends Base {
         await generateThumbnail(buffer, mediaType, options)
         if (mediaType === MessageType.audio && !options.duration) {
             try {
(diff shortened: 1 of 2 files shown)


adiwajshing:
Any reason why this is required?

Borewit:
> Any reason why this is required?

It is not required. Without the providing the MIME-type it will be determined based on it's content.

adiwajshing:
If there is no problem with the current approach, I don't believe we need this PR?

Borewit:
> If there is no problem with the current approach, I don't believe we need this PR?

Correct, if you think the content should determine the type it is fine as it is. If you believe the interpretation should be according to the MIME-type, you should merge. Feel free to close without merging.

Issue #244, opened 2020-11-27
Using lib in a webextension


zoutepopcorn:
This lib looks great, and works fine in node.js

The connection:
```
import WS from 'ws'
```
is working fine in nodejs, but now it's impossible to run in a chrome extension. Maybe it's possible to connect from the extension even without visiting web.whatsapp.com

Hopefully we can also use the browser websocket to connect..

itacirgabral:
I would suggest something like [isomorphic-ws](https://www.npmjs.com/package/isomorphic-ws)

From [ws](https://github.com/websockets/ws) readme:
> Note: This module does not work in the browser. (...). To make the same code work seamlessly on Node.js and the browser, you can use one of the many wrappers available on npm, like isomorphic-ws.

zoutepopcorn:
Ah thanks, that would be great.

I'm sure we need to set the origin header from the websocket now. Tested it without, and failed. The browser is not able to do that. I'm not sure if the chrome extension can do it...

https://stackoverflow.com/questions/22868897/access-websocket-traffic-from-chrome-extension

Or we need to inject it to web.whatsapp.com or just injecting a favicon to connect. I don't know whats possible. If the 'fs' is kept outside the lib, we could save login info in localstorage or something..

itacirgabral:
https://github.com/sigalor/whatsapp-web-reveng#javascript
> just needed because browsers don't allow changing HTTP headers for WebSockets
> (...)
> Also make sure that, when establishing the connection, the HTTP header `Origin: https://web.whatsapp.com` is set, otherwise the connection will be rejected.

https://github.com/sigalor/whatsapp-web-reveng/blob/db7e54755d3fe08f6f34d1d71e42fedbee2c8fd1/index_jsdemo.js#L12
```javascript
new WebSocket("wss://web.whatsapp.com/ws", { headers: { "Origin": "https://web.whatsapp.com" } })
```

zoutepopcorn:
Yes, but this is nodejs I think... In a chrome extension, we have to do a other trick I think

PR #245, opened 2020-11-28
isomorphic websocket

itacirgabral:
https://github.com/adiwajshing/Baileys/issues/244

I tested it against some of my runtimes in backend/node and works fine.
@zoutepopcorn could you test it in your frontend/extension setup?


adiwajshing:
I'm not sure if this would work well on a browser runtime because Baileys has dependencies on other node js libraries like `fs` -- more work might be required to make it completely isomorphic.

itacirgabral:
> more work might be required to make it completely isomorphic.

Browser support may open a lot more use cases. Is it in the Baileys's goals? What do you think about it?

websocket cors:
https://stackoverflow.com/questions/23674199/why-is-there-no-same-origin-policy-for-websockets-why-can-i-connect-to-ws-loc#answer-37837709

Is there node dependencies in the baileys's core?
`Buffer` and `fs` have browserify wrappers.

`qrcode-terminal` and `pino` may have no room on frontend.

adiwajshing:
> Browser support may open a lot more use cases. Is it in the Baileys's goals? What do you think about it?

100% a great thing to have. However, let's have a running version on the browser before merging this PR?

itacirgabral:
I'm using verdaccio to publish the pull requesto to my local registry as `isobaileys`
```
yarn add -g verdaccio
verdaccio &
yarn config set registry http://localhost:4873
```

I've load baileys for create react app:

```typescript
import { useEffect } from 'react';
import { WAConnection } from 'isobaileys';

function App() {
  useEffect(() => {
    const conn = new WAConnection()

    conn.connect().then(() => {
      console.log('connected')
    })
  });

  return (<></>)
}
```

```
{error: TypeError: _a.removeAllListeners is not a function
3.Connect.js:87 WebSocket connection to 'wss://web.whatsapp.com/ws' failed: Error during WebSocket handshake: Unexpected response code: 403
```

[`isomorphic-fetch`](https://github.com/matthew-andrews/isomorphic-fetch) is another easy wrapper

PR #248, opened 2020-11-30
Update example.ts

13banda:
update new message check


adiwajshing:
I think you're introducing a regression?

13banda:
i notice that chat.hasNewMessage no more in v3.3.0

PR #250, opened 2020-11-30
Query update

PouyaAklion:
actually, my problem was loading messages that were sent today in a group so I wrote two new methods at MessagesExtra.ts.

Hoping to be accepted 😊

thank you

### DIFF
--- a/src/WAConnection/7.MessagesExtra.ts
+++ b/src/WAConnection/7.MessagesExtra.ts
@@ -144,6 +144,46 @@ export class WAConnection extends Base {
      * @param chunkSize the number of messages to load in a single request
      * @param mostRecentFirst retreive the most recent message first or retreive from the converation start
      */
+    loadTodayMessages(jid: string, onMessage: (m: WAMessage) => Promise<void>|void, chunkSize = 25, mostRecentFirst = true) {
+        let offsetID = null
+        let stop = false
+        const loadMessage = async () => {
+            const {messages} = await this.loadMessages(jid, chunkSize, offsetID, mostRecentFirst)
+            // callback with most recent message first (descending order of date)
+            let lastMessage
+            if (mostRecentFirst) {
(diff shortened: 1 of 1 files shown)


adiwajshing:
I apologise for not making this clear enough, but Baileys has a generic function to load all messages fitting your criteria. It's called `findMessage` & you can use it like this for your purpose with only a few lines of code:
``` ts

    await conn.connect()

    const isToday = (someDate: Date) => {
        const today = new Date()
        return someDate.getDate() == today.getDate() &&
            someDate.getMonth() == today.getMonth() &&
            someDate.getFullYear() == today.getFullYear()
    }
    const todayMessages = []
    conn.findMessage(
        '1234@s.whatsapp.net', 
        20, 
        msg => isToday(new Date(toNumber(msg.messageTimestamp)*1000)) && todayMessages.push(msg) && true)

```

itacirgabral:
there is `messagesReceivedAfter`

https://github.com/adiwajshing/Baileys/blob/84edfe1ca0b4b5482cd5d923b4d16b8034aba71b/src/WAConnection/7.MessagesExtra.ts#L196-L216

PouyaAklion:
I am using [https://www.npmjs.com/package/baileys](Baileys npm package) it doesn't seem to be the same as Github.

PouyaAklion:
I am using [Baileys npm package](https://www.npmjs.com/package/baileys) it doesn't seem to be the same as Github.

itacirgabral:
> I am using [Baileys npm package](https://www.npmjs.com/package/baileys) it doesn't seem to be the same as Github.

Vanilla hat v1.1.1 https://www.npmjs.com/package/baileys
Red hat v3.3.1 https://www.npmjs.com/package/@adiwajshing/baileys

Could be interesting Baileys have an addon system. Maybe `7.MessagesExtra.ts` could be all in the `@adiwajshing/baileysAddonMessagesExtra` addon.

PR #272, opened 2020-12-12
Improved typedoc page

edgardmessias:
This PR improve the doc page, removindo duplicate entrys.

image

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,6 +14,6 @@
     "declaration": true,
     "lib": ["es2020", "esnext.array"]
   },
-  "include": ["src/*/*.ts"],
+  "include": ["src/**/*.ts"],
   "exclude": ["node_modules", "src/Tests/*", "src/Binary/GenerateStatics.ts"]
 }
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "3.4.0",
   "description": "WhatsApp Web API",
   "homepage": "https://github.com/adiwajshing/Baileys",
-  "main": "lib/WAConnection/WAConnection.js",
-  "types": "lib/WAConnection/WAConnection.d.ts",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
   "keywords": [
     "whatsapp",
     "js-whatsapp",
@@ -46,8 +46,7 @@
     "ws": "^7.3.1"
   },
   "files": [
-    "lib/WAConnection/*",
-    "lib/Binary/*",
+    "lib/*",
     "WAMessage/*"
   ],
   "devDependencies": {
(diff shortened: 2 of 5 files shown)


tsconfig.json
```diff
@@ -14,6 +14,6 @@
     "declaration": true,
     "lib": ["es2020", "esnext.array"]
   },
-  "include": ["src/*/*.ts"],
+  "include": ["src/**/*.ts"],
```
adiwajshing, line 17: Why is this here?

src/typedocEntry.ts
```diff
@@ -0,0 +1,5 @@
+export * from './../WAMessage/WAMessage'
```
adiwajshing, line 1: Doesn't `src/WAConnection/WAConnection.ts` work instead of this?

adiwajshing:
@edgardmessias PR is much appreciated. Have added a few questions.

tsconfig.json
edgardmessias: * is one or more subfolders
** is zero, one or more subfolders

It is necessary to include root files to compile, like src/<name>.ts

src/typedocEntry.ts
edgardmessias: No, because it will generate docs only of exported methods, classes and constants.

The beta version of typedoc works like a library instead of module and each entry is a module, so to generate a single library, we must create a single file with all exportable content.

The idea is the same of files entry in package.json

tsconfig.json
```diff
@@ -14,6 +14,6 @@
...
   "exclude": ["node_modules", "src/Tests/*", "src/Binary/GenerateStatics.ts"]
```
adiwajshing, line 18: Should probably exclude `typedocentry.ts`

edgardmessias: No, because typedoc use the typescript compiler to generate the docs.

As a alternative, we can rename the `typedocentry.ts` to `index.ts` and edit the files of package.json from `"files": ["lib/WAConnection/*","lib/Binary/*","WAMessage/*"]` to `"files": ["lib/index.js"]`

edgardmessias: The ideia is to show what the `@adiwajshing/baileys` exports

adiwajshing: Yeah I'm good with that

edgardmessias: So, @adiwajshing , Would you like me to make these changes and update this PR?

adiwajshing: yes please

edgardmessias: I did

adiwajshing:
@edgardmessias looks good. Just resolve the conflict & I'll merge.

edgardmessias:
@adiwajshing  Done

PR #277, opened 2020-12-14
Fixed event emitter for default callback

edgardmessias:
This PR fix the problem with multiple listeners with same kind of callback when only first is emitted.
Example:
* `CB:action,add:update,message`
* `CB:action,add:update`
* `CB:action,add` // new
* `CB:action,,message`
* `CB:action`

Previous this PR, only `CB:action,add:update,message` is emitted

This PR fix the `CB:Conn,pushname` example in README.md too

To test (it takes a while): `npm install github:edgardmessias/Baileys#fix_emit`

### DIFF
--- a/src/WAConnection/3.Connect.ts
+++ b/src/WAConnection/3.Connect.ts
@@ -202,12 +202,15 @@ export class WAConnection extends Base {
             const l0 = json[0] || ''
             const l1 = typeof json[1] !== 'object' || json[1] === null ? {} : json[1]
             const l2 = ((json[2] || [])[0] || [])[0] || ''
+
             Object.keys(l1).forEach(key => {
-                anyTriggered = anyTriggered || this.emit (`${DEF_CALLBACK_PREFIX}${l0},${key}:${l1[key]},${l2}`, json)
-                anyTriggered = anyTriggered || this.emit (`${DEF_CALLBACK_PREFIX}${l0},${key}:${l1[key]}`, json)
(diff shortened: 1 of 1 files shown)


src/WAConnection/3.Connect.ts
```diff
@@ -202,12 +202,23 @@ export class WAConnection extends Base {
...
             const l1 = typeof json[1] !== 'object' || json[1] === null ? {} : json[1]
             const l2 = ((json[2] || [])[0] || [])[0] || ''
+
+            const triggers = [];
             Object.keys(l1).forEach(key => {
-                anyTriggered = anyTriggered || this.emit (`${DEF_CALLBACK_PREFIX}${l0},${key}:${l1[key]},${l2}`, json)
-                anyTriggered = anyTriggered || this.emit (`${DEF_CALLBACK_PREFIX}${l0},${key}:${l1[key]}`, json)
+                triggers.push(`${DEF_CALLBACK_PREFIX}${l0},${key}:${l1[key]},${l2}`);
+                triggers.push(`${DEF_CALLBACK_PREFIX}${l0},${key}:${l1[key]}`);
+                triggers.push(`${DEF_CALLBACK_PREFIX}${l0},${key}`);
```
adiwajshing, line 210: Instead of creating array, we can just stick to the boolean model. It'll be less resource intensive

edgardmessias: Fixed

PR #279, opened 2020-12-15
Added method to clear chat and star/unstar message

edgardmessias:
* Added method to clear chat
* Added method to star/unstar message

To test (it takes a while): `npm install github:edgardmessias/Baileys#feat_clear_chat`

### DIFF
--- a/src/Tests/Tests.Misc.ts
+++ b/src/Tests/Tests.Misc.ts
@@ -116,6 +116,66 @@ WAConnectionTest('Misc', conn => {
         await delay (2000)
         await conn.modifyChat (testJid, ChatModification.unmute)
     })
+    it('should star/unchar messages', async () => {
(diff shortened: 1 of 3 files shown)


src/Tests/Tests.Misc.ts
```diff
@@ -116,6 +116,11 @@ WAConnectionTest('Misc', conn => {
         await delay (2000)
         await conn.modifyChat (testJid, ChatModification.unmute)
     })
+    it('should clear a chat', async () => {
```
adiwajshing, line 119: Might be a better idea to send messages to a random jid, and then clear that. People might lose a chat with this, otherwise

edgardmessias: I resolved this problem using a chat with yourself

adiwajshing:
@edgardmessias Thanks a lot for this

PR #384, opened 2021-03-01
Add support for contacts array

daganomri:
While working with this repo I noticed that the contactsArrayMessage was already typed but not integrated in the message flow.
This PR adds contactsArrayMessage to the message flow.

### DIFF
--- a/Example/example.ts
+++ b/Example/example.ts
@@ -32,7 +32,7 @@ async function example() {
     // loads the auth file credentials if present
     /*  Note: one can take this auth_info.json file and login again from any computer without having to scan the QR code, 
         and get full access to one's WhatsApp. Despite the convenience, be careful with this file */
-    fs.existsSync('./auth_info.json') && conn.loadAuthInfo('./auth_info.json')
+    fs.existsSync('./auth_info.json') && conn.loadAuthInfo ('./auth_info.json')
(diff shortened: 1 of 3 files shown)


Example/example.ts
```diff
@@ -93,6 +93,10 @@ async function example() {
         } else if (messageType === MessageType.contact) {
             const contact = m.message.contactMessage
             console.log(sender + ' sent contact (' + contact.displayName + '): ' + contact.vcard)
+        } else if (messageType === MessageType.contactsArray) {
+            const contacts = m.message.contactsArrayMessage
+            console.log(sender + ` sent ${contacts.displayName}:`)
+            contacts.contacts.forEach(contact => console.log(`${contact.displayName}: ${contact.vcard}`))
```
adiwajshing, line 99: can remove these logs

daganomri: I removed them.

Example/example.ts
```diff
@@ -32,15 +32,15 @@ async function example() {
     // loads the auth file credentials if present
```
adiwajshing, line 32: Can remove changes from example.ts -- it's only a prettier run on it

daganomri: I removed the prettier formatting.

PR #395, opened 2021-03-12
Add the option to retry infinitely

NilsIrl:


adiwajshing:
@NilsIrl can't you replicate the same behaviour by just setting `conn.connectOptions.maxRetries = Infinity`

NilsIrl:
> @NilsIrl can't you replicate the same behaviour by just setting `conn.connectOptions.maxRetries = Infinity`

I didn't know about `Infinity` in javascript. I don't see any reason why it wouldn't work.

PR #421, opened 2021-03-26
Added method to revokes link of group chat

ibnusyawall:
Added method to revokes link of group chat, 

example:

```js
let response = await client.revokeInvite('abcd-xyz@g.us');

console.log(response);
```

I've Test with great results and works well


RizqiTatsumaki:
harusnya bagi aku, biar ak yang pull

PR #425, opened 2021-03-29
getBusinessProfile()

Nurutomo:


src/WAConnection/5.User.ts
```diff
@@ -211,4 +211,32 @@ export class WAConnection extends Base {
...
         return result
     }
+    /**
+     * Query Business Profile (Useful for VCards)
+     * @param jid Business Jid
+     * @returns profile object or undefined if not business account
+     */
+    async getBusinessProfile(jid: string) {
+        jid = whatsappID(jid)
+        const { isBusiness } = await this.isOnWhatsApp(jid) || { isBusiness: false }
```
adiwajshing, line 221: this might be an unnecessary query, or at least should be optional

PR #507, opened 2021-06-01
📱 Add Video/Image Resolution Metadata using FFprobe

Nurutomo:
## Changes
1. `MessageOptions`:
   - media `width`
   - media `height`

2. `package.json`:
   - `ffprobe` package
   
3. `Utils.ts`:
   - `extractMediaMessage`
   ```ts
   extractMediaMessage(file: string, ffprobePath?: string) => Promise<ffprobe.FFProbeResult>
   ```
   
   
#### WhatsApp Android
WhatsApp Android
#### WhatsApp Web
WhatsApp Web

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ import * as fs from 'fs'
 
 const conn = new WAConnection() 
 // this will be called as soon as the credentials are updated
-conn.on ('credentials-updated', () => {
+conn.on ('open', () => {
     // save credentials whenever updated
     console.log (`credentials updated!`)
     const authInfo = conn.base64EncodedAuthInfo() // get all the auth info we need to restore this session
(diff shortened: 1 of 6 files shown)


jeremia49:
Good PR, but I think you need to also include ffprobe-static , in case the system hasn't installed ffprobe yet.

IMO, I don't think we need this PR , because it's only make small difference. At same time, considering we need to depends on some binary build that may not exists or works in some system. 

I think one can make themselves if they want to. CMIIW

Nurutomo:
Sorry about `0.Base.ts`, I pressed `Alt + Shift + F` there
https://github.com/adiwajshing/Baileys/pull/507/files#diff-ee3806277def7f3d927ae861e384827fbcd3afc829c6b91324fde4cd10c7d24b

Nurutomo:
Changes:

image

Nurutomo:
i dont know how does `streamingSidecar` actually generated

moskoweb:
Why wasn't it merged?

PR #554, opened 2021-06-27
Stream as input

Nurutomo:


AlenSaito1 (changes requested):
Is it necessary to add two extra dependencies?

You can check if an Object is readable stream by the following 
```TS
const isReadableStream = (stream: unknown): boolean =>
	typeof stream.pipe === 'function' &&
	!stream.readable &&
	typeof stream._read === 'function' &&
	typeof stream._readableState === 'object'
```
or 
```TS
import { Readable } from 'stream'

const isReadableStream = (stream: unknown): boolean => stream instanceof Readable
```

anshuljai:
nice bro :D

PR #572, opened 2021-07-05
Slicing extra message returned by WhatsApp

bhavya32:
WhatsApp returned count+1 messages in `fetchMessagesFromWA()`. I sliced it to remove the extra message. 
In case `indexMessage` is specified, it returns count messages only. So no slice in that condition. 
Do note that though I have checked once, I am not sure if it breaks something else.


bhavya32:
Opening another proper pr with 1 more update.

Issue #564, opened 2021-07-02
[BUG] Application crashes due to incorrect ID format when sending messages (Request to update documentation)


gunslingerOP:
**Describe the bug**
sendMessage method crashes WhatsApp application and causes weird behavior, I found out the hard way that one should not enter a "+" before the number they're sending it to

This way made the app crash and messed up everything : ["+" then country code][Phone number]

The correct way of making an ID is : [country code WITHOUT a +][Phone number you're sending to]@s.whatsapp.net

**To Reproduce**
1. Created a new connection and authenticated it
2. try to send a normal text message
3. Provide the ID parameter in this format :  ["+" then country code][Phone number]
4. Whatsapp app on your phone crashes and your node app will keep trying to find a phone connection

**FIX**
Simply **do not** add a + before the country code when creating an ID, it should be created in this format:

[country code WITHOUT a +][Phone number you're sending to]@s.whatsapp.net

Maintainers, please update the documentation as you explicitly show a + before the country code!

bhavya32:
@gunslingerOP You are absolutely right. I will make PR.

Issue #559, opened 2021-06-29
isOnWhatsAppNoConn still works? Android app crashing!


canove:
I have a number that once I send a message to it, it crashes de Android app
The number is 000000000006, you can give a try (be careful, it will crash the mobile app)

I'm working in a way to check if the number exists on my contact creation logic. Since I don't always have a connection available, I tried to use isOnWhatsAppNoConn, but it always return `{ exists: true, jid: '000000000006@s.whatsapp.net' }`

- This method still works for someone else? If it's not, I suggest to remove it from the isOnWhatsApp fallback, since it can lead to errors.
- Is there a way to check if a number exists without a connection?

khoirulh1610:
I've experienced something similar, it happened because of my mistake in the script.

gunslingerOP:
> I've experienced something similar, it happened because of my mistake in the script.

@khoirulh1610 Could you say what the mistake was? I'm facing a similar issue when trying to send a message!

bhavya32:
@gunslingerOP I can confirm its broken now.

bhavya32:
@canove I can confirm `isOnWhatsAppNoConn` is broken and I proposed to remove it as a fallback. Currently there is no other method. There is no need for connection between phone and baileys for `isOnWhatsApp` . You just have to connect once and it should then work even if phone is not connected to internet. (Not Sure about it. Tested for like 30 min)

PR #573, opened 2021-07-05
Fixed: loadMessages And isOnWhatsApp and Updated Readme

bhavya32:
1. `loadMessages` - WA returned count+1 messages in case `indexMessage `  is not specified. This caused wrong number of elements in `loadMessages` and `loadUnreadMessages`. This should fix it.

2. `isOnWhatsAppNoConn` is now broken as the API it relied upon simply does not check if the number is on WhatsApp. So remove it from fallback to avoid false positives. This will now result in an error in case connection is not open. Found in #559.

3. Update Readme as noticed by #564.

### DIFF
--- a/src/WAConnection/7.MessagesExtra.ts
+++ b/src/WAConnection/7.MessagesExtra.ts
@@ -93,6 +93,9 @@ export class WAConnection extends Base {
             null,
         ]
         const response = await this.query({json, binaryTags: [WAMetric.queryMessages, WAFlag.ignore], expect200: false, requiresPhoneConnection: true})
+		if(!indexMessage&&response[2]){
+			response[2] = response[2].slice(1)
+		}
         return (response[2] as WANode[])?.map(item => item[2] as WAMessage) || []
     }
     /**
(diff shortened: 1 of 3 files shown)


bhavya32:
@adiwajshing It would be nice if you close the PR in case there is any problem with it instead of staling it. Feels like ghosting.

src/WAConnection/7.MessagesExtra.ts
```diff
@@ -93,6 +93,9 @@ export class WAConnection extends Base {
             null,
         ]
         const response = await this.query({json, binaryTags: [WAMetric.queryMessages, WAFlag.ignore], expect200: false, requiresPhoneConnection: true})
+		if(!indexMessage&&response[2]){
```
adiwajshing, line 96: @bhavya32 I don't know how this will react with already existing code out there. Maybe it's best to remove this to maintain consistency? The rest is good to go

bhavya32: @adiwajshing I tested LoadMessages and LoadAllMessages multiple times and it was working fine. If you still hesist, please go ahead without it.

adiwajshing: @bhavya32 yeah please remove it, we don't have a test for this -- I don't know what will happen

bhavya32: @adiwajshing removed it.

Issue #488, opened 2021-05-13
Marker was Not Found! [Get Media with Axios, Cheerio etc..]


phaticusthiccy:
As we know, Baileys has updated new version to 3.5.0
The update is very good overall, but there are some errors like this issue. Called "**marker was not found**".
This error occurs from any link and when writing a non-local file. I searched and found that the problem was caused by the thumbnail. I'll show you an alternative route for pass this error.
Firstly it's working perfectly but it will not be the same as before. I mean auto thumbnails in media messages.
Step by step:

##

### `Step 1 - Upload a Thumbnail Photo`
Firstly we must upload a photo inside our local files.
I am creating new folder name **media**

![IMG_20210513_202054](https://user-images.githubusercontent.com/74794033/118161861-c1ea9580-b428-11eb-9490-d0ce89864c9a.jpg)
(`mkdir media`)

Then we must upload any thumb inside this folder.

![IMG_20210513_202747](https://user-images.githubusercontent.com/74794033/118162693-b64b9e80-b429-11eb-80b6-d922b226c6d4.jpg)
(`cd media/`) - (`cp {thumb way} {media folder way}`)

### `Step 2 - Adding Function Inside Code`

Now, you must add **↓ this ↓** function inside your code.
```js
const fs = require('fs')
function base64_encode(file) {
    var bitmap = fs.readFileSync(file);
    return new Buffer(bitmap).toString('base64');
}
var base64str = base64_encode("{new thum way}");
```
Here is an example:

![PSX_20210513_203918](https://user-images.githubusercontent.com/74794033/118164049-548c3400-b42b-11eb-9191-2cf5e62d8f77.jpg)

##

### `Step 3 - Add Thumbnail inside Media Messages`

Now after this time, you must use your own structure. We are going to use thumbnail from [here](https://github.com/adiwajshing/Baileys#notes)
When sending media messages, you must use the following code: 
```js
{ **thumbnail: base64str**, mimetype: Mimetype.jpeg, etc.. }
```
Here is an example:

![IMG_20210513_204659](https://user-images.githubusercontent.com/74794033/118164974-64f0de80-b42c-11eb-952a-fd4c1616e8ee.jpg)

##

This is the simple solution to the "marker was not found" error. You can write your comments and suggestions under this issue :)

Special Thanks @adiwajshing for codded Bailyes :)

eduardohf14:
@phaticusthiccy 
[...]

PR #639, opened 2021-08-11
Wait for drain event before calling Writable.write

saviosg:
This is my solution for the bug that caused the program to crash when you try to send certain kinds of images.
After investigating the issue I found it was likely due to some kind of race condition related to buffering as discussed in #488 
Not 100% sure why this is happening, but I changed the writable stream code to do as recommended in Node's documentation and now it's working correctly.
I'm sending this PR because it seems this issue is still around.

### DIFF
--- a/src/WAConnection/Utils.ts
+++ b/src/WAConnection/Utils.ts
@@ -14,6 +14,7 @@ import KeyedDB from '@adiwajshing/keyed-db'
 import got, { Options, Response } from 'got'
 import { join } from 'path'
 import { IAudioMetadata } from 'music-metadata'
+import { once } from 'events'
 
 const platformMap = {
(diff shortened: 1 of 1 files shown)


phaticusthiccy:
#488 Merge still don't work for jpeg/jpg files. Just a little reminding.

saviosg:
What I mean by "reproduce" is, are you able to run into this bug if you run the code from the master branch? 

As for the function compressImage, I have also considered the possibility of it being the source of the bug, but I didn't find anything in it that could be causing the bug.

phaticusthiccy:
Gotcha, yeah I am getting this error on master branch. And maybe your right but this error only appears when users trying to send jpeg or jpg images. I don't know why it is caused but its so annoying.

PR #644, opened 2021-08-13
fix post story text backgroundArgb missing after generatedLinkPreview proses

hasanbasri1993:


hasanbasri1993:
how about this pr @PurpShell ?

PurpShell:
Does this even generate the preview? Please try normally not on story

PR #650, opened 2021-08-17
fix post story text backgroundArgb missing after generatedLinkPreview

hasanbasri1993:
@PurpShell 

Yes, [commit before](https://github.com/hasanbasri1993/Baileys/commit/b53a82f8e9959db5b7d1b0b214c92b6b379cf056), remove generated Preview link
But. this commit work well

image

image


src/WAConnection/6.MessagesSend.ts
```diff
@@ -447,14 +448,12 @@ export class WAConnection extends Base {
         return trueFileName
     }
     /** Query a string to check if it has a url, if it does, return required extended text message */
-    async generateLinkPreview (text: string) {
-        const query = ['query', {type: 'url', url: text, epoch: this.msgCount.toString()}, null]
+    async generateLinkPreview(message) {
```
adiwajshing, line 451: @hasanbasri1993 this is a big, breaking change. I don't see why it's there

hasanbasri1993: Actually when post status@broadcast when add  backgroundArgb or font 
After this method them lost 

to prevent lost so try change parameter text: string only into message

adiwajshing: Not sure if a breaking change is a good idea for something like this

hasanbasri1993: image


i tried to run Test SendMessage only, send linkpreview is passed

adiwajshing: It's not a breaking change there, it is a breaking change if someone is using `generateLinkPreview` directly

hasanbasri1993: so any solution breaking change if someone is using it ?

src/WAConnection/6.MessagesSend.ts
```diff
@@ -18,6 +18,8 @@ import {
 import { isGroupID, generateMessageID, extensionForMediaMessage, whatsappID, unixTimestampSeconds, getAudioDuration, newMessagesDB, encryptedStream, decryptMediaMessageBuffer, generateThumbnail  } from './Utils'
 import { Mutex } from './Mutex'
 import { Readable } from 'stream'
+import {proto} from "../../WAMessage/WAMessage";
+import ButtonsMessage = proto.ButtonsMessage;
```
ibnusyawall, line 22: ButtonsMessage for what?

hasanbasri1993: See the last commit, in second commit i have remove

src/WAConnection/6.MessagesSend.ts
adiwajshing: Yes, of course

src/WAConnection/6.MessagesSend.ts
adiwajshing: It's still there

hasanbasri1993: ok i have removed it, sorry i haven't  stage into commit

PR #658, opened 2021-08-19
Fix send contacts

YehudaEi:
There is a problem sending contacts, and this fix will fix it.

(It is very possible that this fault also exists in other types of messages where this function is not static).


adiwajshing:
@YehudaEi your fix edits an auto-generated file, it's not going to fix any issue

1. if there is a bug in sending contacts, describe it, attach more info -- we can't go by vague descriptions
2. if there is a bug when sending other types of messages -- again, describe the issue

YehudaEi:
@adiwajshing 
Sorry, I just do not speak typescript ...

My problem occurred while trying to run the example that appears in the readme and I received an error regarding this function which is private.

WAMessage/WAMessage.d.ts
```diff
@@ -9548,7 +9548,7 @@ export namespace proto {
          * Converts this ContactMessage to JSON.
          * @returns JSON object
          */
-        public toJSON(): { [k: string]: any };
+        public static toJSON(): { [k: string]: any };
```
ibnusyawall, line 9551: public static? Which one is true?

YehudaEi: My mistake - I meant it should be public static (and not an instance method).

YehudaEi:
Hi @adiwajshing,
Did you understand my problem?

adiwajshing:
Not really @YehudaEi

YehudaEi:
What I was trying to do is run the example of sending a contact brought in readme.
I did nothing beyond that, and then I got an error (screenshot attached).
The way I solved the error was with adding a static before the function.

image
image

YehudaEi:
Hi @adiwajshing,
Did you understand my problem now?

YehudaEi:
Hi @adiwajshing,
Did you understand my problem now?

PR #699, opened 2021-09-23
[multi device] fix install for npm package and some typo fix

pokearaujo:


adiwajshing:
@pokearaujo thanks for the PR, I didn't realise you cannot install the package -- I'll fix up the store and update, there are some more changes I need to make the store function & write the docs.

PR #708, opened 2021-09-29
Added Context Info Option

viniciusgdr:
Added the Option Context Info in MiscMessageGenerationOptions.


adiwajshing:
Not sure what the purpose for this is?

viniciusgdr:
> Not sure what the purpose for this is?

In the documentation it says that it has this option, but when I entered the /src/Types/Message.ts file the option was not available. In the case there I added what was missing.

PR #804, opened 2021-11-07
Fixed: Update for new Chat Groups

viniciusgdr:
The WhatsApp changed the form of Group ID, so it no longer has the '-'

### DIFF
--- a/src/WAConnection/4.Events.ts
+++ b/src/WAConnection/4.Events.ts
@@ -494,10 +494,6 @@ export class WAConnection extends Base {
     /** Adds the given message to the appropriate chat, if the chat doesn't exist, it is created */
     protected async chatAddMessageAppropriate (message: WAMessage) {
         const jid = whatsappID(message.key.remoteJid)
-        if(isGroupID(jid) && !jid.includes('@g.us' || '-')) {
-            this.logger.warn({ gid: jid }, 'recieved odd group ID')
-            return
-        }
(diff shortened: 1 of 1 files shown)


viniciusgdr:
Now the Group ID no longer has the '-', as presented below:

WARN [2021-11-07 01:41:59.019 +0000] (2780): recieved read update for unknown chat 120363039091387898@g.us
    class: "Baileys"

azudindaem:
i created group using md, still having - between id & time.. weird

viniciusgdr:
> i created group using md, still having - between id & time.. weird


So it means that we have 2 types of creation IDs for the new groups

src/WAConnection/4.Events.ts
```diff
@@ -494,7 +494,7 @@ export class WAConnection extends Base {
     /** Adds the given message to the appropriate chat, if the chat doesn't exist, it is created */
     protected async chatAddMessageAppropriate (message: WAMessage) {
         const jid = whatsappID(message.key.remoteJid)
-        if(isGroupID(jid) && !jid.includes('-')) {
```
adiwajshing, line 497: Might be a good idea to remove this check altogether

viniciusgdr: Ok

viniciusgdr: I also put the PR of pe77 because if I closed my PR the his modification would probably be removed

PR #816, opened 2021-11-09
Update README.md

azudindaem:
Adding example on how to send button message


moskoweb:
Is it possible to send buttons that are URL links?

PR #824, opened 2021-11-10
Fixed odd jid of new groups

souravkl11:


viniciusgdr:
there is already a pull pending on this... #804 .

PR #833, opened 2021-11-11
Fix retry receipt, fix assertingPreKeys, uploadPreKeys to 30, generate QR timeout and update example

pokearaujo:


src/Socket/socket.ts
```diff
@@ -174,22 +175,22 @@ export const makeSocket = ({
...
 
+        const preKeys = await getPreKeys(authState.keys, preKeysRange[0], preKeysRange[0] + preKeysRange[1])
+        await execute(preKeys)
+
         ev.emit('auth-state.update', authState)
     }
     /** generates and uploads a set of pre-keys */
     const uploadPreKeys = async() => {
-        await assertingPreKeys(50, async preKeys => {
+        await assertingPreKeys(30, async preKeys => {
```
adiwajshing, line 193: Any reason for not using `50`?

Example/example.ts
```diff
@@ -4,7 +4,7 @@ import { Boom } from "@hapi/boom"
 import makeWASocket, { WASocket, AuthenticationState, DisconnectReason, AnyMessageContent, BufferJSON, initInMemoryKeyStore, delay } from '../src'
 
 (async() => {
-    let sock: WASocket | undefined = undefined
+    let currentSock: WASocket | undefined = undefined
```
adiwajshing, line 7: can we keep the name the same?

src/Socket/messages-recv.ts
```diff
@@ -43,8 +43,15 @@ export const makeMessagesRecvSocket = (config: SocketConfig) => {
         await sendNode(stanza)
     }
 
+    const retries = new Map<string, number>();
     const sendRetryRequest = async(node: BinaryNode) => {
-        const retryCount = +(node.attrs.retryCount || 0) + 1
+        if (retries.has(node.attrs.id) && retries.get(node.attrs.id)! >= 50) {
```
adiwajshing, line 48: Is this necessary?

src/Socket/socket.ts
```diff
@@ -174,22 +175,22 @@ export const makeSocket = ({
...
-        
         creds.serverHasPreKeys = true
         creds.nextPreKeyId = Math.max(lastPreKeyId+1, creds.nextPreKeyId)
         creds.firstUnuploadedPreKeyId = Math.max(creds.firstUnuploadedPreKeyId, lastPreKeyId+1)
         await Promise.all(
             Object.keys(newPreKeys).map(k => authState.keys.setPreKey(+k, newPreKeys[+k]))
         )
 
+        const preKeys = await getPreKeys(authState.keys, preKeysRange[0], preKeysRange[0] + preKeysRange[1])
+        await execute(preKeys)
```
adiwajshing, line 187: Not sure what shifting it here does? Wouldn't it just be uploading 0 pre-keys always?

src/Socket/socket.ts
pokearaujo: In wa web is a 30 pre key bundle, idk if its okay send more

image

src/Socket/socket.ts
```diff
@@ -415,20 +415,39 @@ export const makeSocket = ({
...
+        }
+
+        await genPairQR()
     })
     // device paired for the first time
     // if device pairs successfully, the server asks to restart the connection
     ws.on('CB:iq,,pair-success', async(stanza: BinaryNode) => {
         logger.debug('pair success recv')
         try {
+            if (qrTimer) {
```
adiwajshing, line 447: might be better to put this with `connection.update` and clear it when connection is 'open' or 'close'

src/Socket/messages-recv.ts
pokearaujo: there a max retry
image

but is 5 attempts
image

sometimes don't have retryCount in msg

src/Socket/socket.ts
pokearaujo: in sendRetryReceipt was already coming with 0 prekeys when necessary, but is needed a new prekey to send in retry, and the limit in getPreKeys was wrong, he doesn't enter the for condition (ex: min 30, limit: 1)

image

Example/example.ts
```diff
@@ -32,6 +32,20 @@ import makeWASocket, { WASocket, AuthenticationState, DisconnectReason, AnyMessa
             JSON.stringify(state, BufferJSON.replacer, 2)
         )
     }
+
+    const updateConnection = (update: Partial<ConnectionState>) => {
```
adiwajshing, line 36: can we just keep this in the `connection.update` listener?

src/Socket/socket.ts
```diff
@@ -415,14 +415,43 @@ export const makeSocket = ({
...
         const noiseKeyB64 = Buffer.from(creds.noiseKey.public).toString('base64');
         const identityKeyB64 = Buffer.from(creds.signedIdentityKey.public).toString('base64')
         const advB64 = creds.advSecretKey
-        const qr = [refs[0], noiseKeyB64, identityKeyB64, advB64].join(',');
 
-        ev.emit('connection.update', { qr })
-        await postQR()
-        await sendNode(iq)
+        const genPairQR = async() => {
+            const ms = refs.length == 6 ? 6e4 : 2e4
```
adiwajshing, line 425: this is meant to have a different timeout on the first QR and then for subsequent QRs, right?

If that is the case, may we just keep a boolean or counter for the same -- because if refs length change on WA's side -- this code breaks. 

Also can we use human readable numbers instead of `6e4` or `2e4`?

src/Socket/socket.ts
```diff
@@ -415,14 +415,43 @@ export const makeSocket = ({
...
+        const checkConnection = ({ connection }: ConnectionState) => {
+            if (!['open', 'close'].includes(connection)) {
+                return;
+            }
+
+            if (qrTimer) {
+                clearTimeout(qrTimer)
+            }
+
+            ev.off('connection.update', checkConnection);
```
adiwajshing, line 451: May I suggest:

```
if(connection === 'open' || connection === 'close') {
  clearTimeout(qrTimer)
  ev.off('connection.update', checkConnection)
}
```

src/Socket/socket.ts
```diff
@@ -415,14 +415,43 @@ export const makeSocket = ({
...
+
+            const qr = [ref, noiseKeyB64, identityKeyB64, advB64].join(',')
+    
+            ev.emit('connection.update', { qr })
+            await postQR(qr)
+
+            qrTimer = setTimeout(genPairQR, ms)
+        }
+
+        await genPairQR()
```
adiwajshing, line 440: don't need 'await' here.

src/Socket/socket.ts
```diff
@@ -415,14 +415,43 @@ export const makeSocket = ({
...
+            const ref = refs.shift()
+            if(!ref) {
+                end(new Boom('QR refs attempts ended', { statusCode: DisconnectReason.restartRequired }))
+                return
+            }
+
+            const qr = [ref, noiseKeyB64, identityKeyB64, advB64].join(',')
+    
+            ev.emit('connection.update', { qr })
+            await postQR(qr)
```
adiwajshing, line 435: can remove the await here as well and make this a sync function

Example/example.ts
pokearaujo: after first restart, he will not trigger anymore, because he is listening to the old sock

adiwajshing: I understand -- I meant can this function just be directly put in `sock.ev.on` (line 73)

adiwajshing:
Looks good overall, can remove all semis -- can merge after that @pokearaujo

Example/example.ts
```diff
@@ -50,15 +51,32 @@ import makeWASocket, { WASocket, AuthenticationState, DisconnectReason, AnyMessa
...
+        // listen for when the auth state is updated
+        // it is imperative you save this data, it affects the signing keys you need to have conversations
+        sock.ev.on('auth-state.update', () => saveState())
+
         return sock
     }
 
     const sendMessageWTyping = async(msg: AnyMessageContent, jid: string) => {
-
         await sock.presenceSubscribe(jid)
```
MihailovJava, line 80: what about state of this instance of socket after moving  reconnect inside fun startSock() ?

adiwajshing: It'll attach to the correct reference

src/Socket/messages-recv.ts
```diff
@@ -43,8 +43,15 @@ export const makeMessagesRecvSocket = (config: SocketConfig) => {
         await sendNode(stanza)
     }
 
+    const retries = new Map<string, number>()
     const sendRetryRequest = async(node: BinaryNode) => {
-        const retryCount = +(node.attrs.retryCount || 0) + 1
+        if (retries.has(node.attrs.id) && retries.get(node.attrs.id)! >= 5) {
```
adiwajshing, line 48: I think we can remove the whole map thing -- it's a whole additional state to keep track of -- if it's required later, can add it -- but right now, let's rely on the `retryCount` property

burstfreeze:
@adiwajshing are you still thinking about it or is there anything else you want to see in this PR? I'm happy to do whatever you you need to get this merged since I think it is an important PR

adiwajshing:
Just the retryCount thing, otherwise can merge

Example/example.ts
burstfreeze: > what about state of this instance of socket after moving reconnect inside fun startSock() ?

@MihailovJava do you mean whether after that the `sock` that is being returned from the function can still be accessed after it's been re-assigned, like whether it's a potential value vs reference issue?

burstfreeze:
Made a somewhat clean PR now with reference to the changed commit at https://github.com/adiwajshing/Baileys/pull/848

burstfreeze:
My PR is trash, it messes up everything.

In other words, we NEED the map  (or anything that does the same) in this PR here. I have rolled out to production and the PR even fixed the sessions that broke with my bad non-map PR. I cannot provide long-term testing yet (obviously) but at the moment it looks like Diego made it run perfectly now.

If you think the map should be replaced with something else that is in your judgement but my PR that removed it led to nothing but problems.

Will report back whether it stayed stable

HUGEIT:
I can see many advantages with #833 however just now I have experienced a new combination of issues: I have seen tests on the master branch and I would also love to write tests but ran into the following:

**[ P1 ]** and **[ P2 ]** are phones,
**[ Bot 1.A ]** and **[ Bot 1.B ]** are connected Baileys instances connected to **[ P1 ]**
**[ Bot 2.A ]** is connected to **[ P2 ]**


**[ Bot 1.A ]** sent a message in a group chat. the message did not come onto messages.upsert of **[ Bot 1.A ]** itself

**[ P2 ]** however did receive that message but neither the remaining bots **[ Bot 1.B ]** nor **[ Bot 2.A ]** did receive that message.

now the most confusing part is that even **[ P1 ]** itself cannot see the message.

other messages came in normally I think but this one stuck out.


in other words the visibility table for the content that [ Bot 1.A ] has sent :

```
❌ P1       (sender host)
✅ P2       (group member)
❌ Bot 1.A  (sending bot)
❌ Bot 1.B  (other sending both)
❌ Bot 2.A  (group member bot)
```

so the only database where the message actually appeared is the database of the other phone. not any baileys, not host phone, just that phone

maybe it is an issue to have 2 baileys connecting to each other because it only knows the key exchange with browser so then key exchange goes bad? or 2 baileys on one phone? but it would be nice to write tests to validate usage

adiwajshing:
@burstfreeze thanks for the testing, in that case -- I'll merge the PR as is

adiwajshing:
@HUGEIT so I think your issue is not with the PR, but with the consistency of signal sessions -- it looks like they're messed up, causing the decryption errors you're seeing

PR #846, opened 2021-11-12
pokearaujo PR: fix assertingPreKeys, uploadPreKeys to 30, generate QR timeout and update example

burstfreeze:
Basically pokearaujo's PR #833 but reverted retry map because Adi did not like if I understood correctly.


burstfreeze:
ah mistake nvm, brb

Issue #834, opened 2021-11-11
[BUG] [MD] resyncAppState calling before receive appStateSyncKeyShare in first connection


pokearaujo:
**Describe the bug**
resyncAppState calling before receive appStateSyncKeyShare in first connection, its seen a racing, sometimes came keys before, sometimes not.

**To Reproduce**
1. Create a new connection
2. Probaly will throw "Error: failed to find key \"XXXXXX\" to decode mutation"
3. If not, clear auth, logout and try again

**Expected behavior**
Maybe need wait the keys before request sync collections? Or save in a queue and process after receive the keys

adiwajshing:
Not an issue even if it errors out, doesn't affect the connection

ilyaborodin:
@adiwajshing I think it can affect. Look.

After scanning QR code on my phone, the loader hangs for a long time (about 40-60 seconds) and after that I get an error on my phone that it is not possible to scan qr. Screenshots are attached.

![IMAGE 2021-11-12 13:42:13](https://user-images.githubusercontent.com/48887908/141454170-8f0b89e8-5d30-4cab-b15e-9509f84729a3.jpg)

![IMAGE 2021-11-12 13:42:01](https://user-images.githubusercontent.com/48887908/141454134-dbd91c13-4eaf-4dee-a239-4eb8c90b132d.jpg)


In the terminal, I see that the status "open" comes to me, but the connection is not displayed on my phone. After restarting the connection, the status "open" is also received.

![IMAGE 2021-11-12 13:46:20](https://user-images.githubusercontent.com/48887908/141454687-e341cdde-4922-4613-a3a2-85b80a81dcc0.jpg)

The only problem I see in the terminal is this error, presented below. 
`
{"level":30,"time":1636712914712,"pid":6280,"hostname":"MacBook-Air-Ilya.local","trace":"Error: failed to find key \"AAAAAKmL\" to decode mutation\n    at decodePatches (/Users/ilyaborodin/WebstormProjects/whatsapp_api_md/node_modules/@adiwajshing/baileys-md/lib/Utils/chat-utils.js:275:23)\n    at processTicksAndRejections (internal/process/task_queues.js:95:5)\n    at async resyncAppState (/Users/ilyaborodin/WebstormProjects/whatsapp_api_md/node_modules/@adiwajshing/baileys-md/lib/Socket/chats.js:180:59)","msg":"failed to sync app state"}
`

It happens by accident. But most of the times, I cannot scan the QR code.

[...]

Issue #808, opened 2021-11-08
[MD] Issue after recent commits: Error: failed to find key "AAAAAB2j" to decode mutation


azudindaem:
New error after using latest commit

{"level":30,"time":1636368107963,"pid":7532,"hostname":"localhost","class":"baileys","msg":"recv all pending notifications"}
============================Connection Update============================
Connection Update { receivedPendingNotifications: true } 26
Error: failed to find key "AAAAAB2j" to decode mutation
    at decodePatches (/home/webapps/md/node_modules/@adiwajshing/baileys-md/lib/Utils/chat-utils.js:275:23)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async resyncAppState (/home/webapps/md/node_modules/@adiwajshing/baileys-md/lib/Socket/chats.js:180:59) {
  data: null,
  isBoom: true,
  isServer: true,
  output: {
    statusCode: 500,
    payload: {
      statusCode: 500,
      error: 'Internal Server Error',
      message: 'An internal server error occurred'
    },
    headers: {}
  }
}

burstfreeze:
I also had that with an older commit, may not be entirely related to the most recent commits. For me it was a chat being pinned but I lost the session before I could open an issue.

If this only happens with pins, I would suggest not breaking the connection based on that

azudindaem:
i see. it just happen when i update to the latest commit just now..

burstfreeze:
Ok update - it just happened to me after scanning the code. Will try again with a clean fork but indeed appears to be an issue.

burstfreeze:
Also happens with a clean fork. @azudindaem please change the empty title of the issue to something like

`Issue after recent commits: Error: failed to find key "AAAAAB2j" to decode mutation`

canove:
Same here.

It just occur on first connection (right after reading QR). 

``` 
[1636372000947] INFO: requesting account sync
    fromTimestamp: "1636372000"
(node:69858) UnhandledPromiseRejectionWarning: Error: failed to find key "AAAAAA8U" to decode mutation
    at Object.decodePatches (/home/cassio/code_learn/whaticket-closed/wpp-md-microservice/src/libs/baileys-md/lib/Utils/chat-utils.js:275:23)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
[...]

PR #848, opened 2021-11-12
#833 update to revert retryCount change

burstfreeze:
For future reference I mentioned the fixed commit in my commit message as Adi said maybe we need this later.

Basically the block should look like it did before now but I haven't had the chance to cause a retry and therefore test it yet.


azudindaem:
It looks good

burstfreeze:
Hey again need to self close. There is an issue in it that results in an infinite request loop for some reason.

@pokearaujo could you check whether I made a mistake reverting the one commit? Like potentially not respecting a maximum you set somewhere else in the lines I removed or something? And if so, is there a potential issue in doing the retry at all since it seems to fail entirely once it fails for the first time?


Log:

```
"sent retry receipt"}
saving pre-keys
saving pre-keys
"fromMe":false,"frame":{"tag":"ack","attrs":{"from":"123@g.us","class":"receipt","type":"retry","id":"123","participant":"123@s.whatsapp.net"},"content":null},"msg":"communication"}
"fromMe":false,"frame":{"tag":"message","attrs":{"from":"123","type":"text","id":"123","participant":"123","t":"1636834002","notify":"123"},"content":[{"tag":"enc","attrs":{"count":"1","v":"2","type":"msg"},"content":{"type":"Buffer","data":[like 300 bytes]}}]},"msg":"communication"}
"trace":"SessionError: No session record\n    at Object.123.0 [as awaitable] (/libsignal/src/session_cipher.js:167:23)\n    ...,"msg":"failure in decrypting message"}
"sent retry receipt"}

```

Then sometimes a message with about 8000 bytes in content buffer data (yes you read that correctly) which also throws the session record.

After restarting the session we see the following anatomy:

`critical_block`, kilobytes of buffer data

`failed to find key \"AAA..\" to decode mutation`

`failed to sync app state`

A lot of incoming receipt requests

Receiving encrypted messages with buffer printing to console (kinda looks like history sync, as if it's re-sending everything)

A lot of failure in decrypting again, supposedly one for each message

A lot of ```{"level":30,"time":123,"pid":2063,"hostname":"","msgId":"123","retryCount":1,"msg":"sent retry receipt"}
saving pre-keys``` supposedly for each message.

A lot of `saving pre-keys` prints after that

A lot of receipt retries

Then a loop from everything in the first half of this comment here like topkek described


I suppose it's #808 again or your #834 except that the immediate launch is fine, but reconnects after that cause issues.




Any ideas?

burstfreeze:
Steps for me to reliably reproduce:

Checkout this pull head. Scan code, let the sync finish. Terminate the session, launch again. After a few seconds you will spot a `critical_block` message with a lot of data, then `failed to find key \"A.." to decode mutation, failed to sync app state` and then the fun begins.

PR #851, opened 2021-11-13
Added new features to GroupParticipantsUpdate

viniciusgdr:
Added: Recognition of the Actor in Promote/Demote/Remove
Added: Action 'leave'


viniciusgdr:
> You cannot use baileys master soon anymore. You need Baileys MD

Yes, I am aware that the connection will change. Here my 2 WhatsApps are already with Baileys MD

PR #863, opened 2021-11-15
Multi device - Fix: groupParticipantsUpdate return list with status code

akhilnarayanan1:
The **groupParticipantsUpdate** is only returning the ID of participants that got affected (without any status code)
image

This fix should return **groupParticipantsUpdate** with error status code (if any)
image


src/Socket/groups.ts
```diff
@@ -98,7 +98,7 @@ export const makeGroupsSocket = (config: SocketConfig) => {
 			)
 			const node = getBinaryNodeChild(result, action)
 			const participantsAffected = getBinaryNodeChildren(node!, 'participant')
-			return participantsAffected.map(p => p.attrs.jid)
+			return participantsAffected.map(p => p.attrs)
```
adiwajshing, line 101: Need TS type for this

Issue #874, opened 2021-11-16
[BUG]  TypeError: Cannot read properties of undefined (reading 'key')


lyfe00011:
Nov 16 23:02:23 Connecting...
Nov 16 23:02:23 {
Nov 16 23:02:23   connection: 'connecting',
Nov 16 23:02:23   receivedPendingNotifications: false,
Nov 16 23:02:23   qr: undefined
Nov 16 23:02:23 }
Nov 16 23:02:24 Connecting...
Nov 16 23:02:24 { connection: 'open' }
Nov 16 23:02:24 Connected
Nov 16 23:02:24 /root/LyFE/node_modules/@adiwajshing/baileys-md/lib/Socket/messages-recv.js:447
Nov 16 23:02:24         const chat = { id: messages[0].key.remoteJid };
Nov 16 23:02:24                                        ^
Nov 16 23:02:24 TypeError: Cannot read properties of undefined (reading 'key')
Nov 16 23:02:24     at EventEmitter.<anonymous> (/root/LyFE/node_modules/@adiwajshing/baileys-md/lib/Socket/messages-recv.js:447:40)
Nov 16 23:02:24     at EventEmitter.emit (node:events:402:35)
Nov 16 23:02:24     at WebSocket.<anonymous> (/root/LyFE/node_modules/@adiwajshing/baileys-md/lib/Socket/messages-recv.js:383:12)
Nov 16 23:02:24     at processTicksAndRejections (node:internal/process/task_queues:96:5)
Nov 16 23:02:24 Tue Nov 16 2021 23:02:24 GMT+0530 (India Standard Time)
Nov 16 23:02:24 Program node test.js exited with code 1
Nov 16 23:02:24 Starting child process with 'node test.js'
after some hours(5+) this error came 
also it run without any restart or connection issues upto this.
yup latest commit.

canove:
Same here. How did you solve it?

ilyaborodin:
Same

ilyaborodin:
At the same time with "Cannot read properties of undefined" I received this errors:
<img width="1440" alt="Снимок экрана 2021-11-17 в 19 34 32" src="https://user-images.githubusercontent.com/48887908/142242606-db959165-f99a-4ff5-b9b2-c0f8dd2729c3.png">

lyfe00011:
> Same

?.key?.remoteJID
im trying this, i thinks happen just after received notification true

ilyaborodin:
> > Same
> 
[...]

Issue #870, opened 2021-11-15
[BUG] "Waiting for this message. This may take a while." in both directions


jackcokk:
Amazing work adiwajshing and pokearaujo and all contributors, we all have to thank you so much!

I am experiencing an issue with messages sent with Baileys that in the message frame it says 

_Waiting for this message. This may take a while._ on a recipients' device.

My sql also sometimes does not have all the messages sent in a chat.

On the other hand in some of these cases, other group participants, my WhatsApp Web desktop app and even another Baileys instance did receive and decrypt that message. Just not the phone itself or sometimes it is available on the phone but not in Baileys.

I suppose it is not an issue with the sending of messages but rather with the key back-and-forth. I can only assume that WAWeb writes these missing messages to a queue and regularly checks whether both the devices in question are online at the same time in order to trigger a new key synchronization.*

To summarize in a user story:

Baileys sends message, someone's device (iPhone or linked device) is offline. The device in question is not in sync with the signal keys (?) and when it goes online, power at the Baileys computer cuts off and is unavailable for a few hours. All devices are online again after a while but the messages from and to each other may be missing.

I have even seen the same among the actual apps but usually within seconds the issue has resolved itself when both went online. So I guess there must be a way to do it in the same manner the website app does it.

Is there something we can do with the website app to e.g. log incoming and outgoing traffic in order that may help you analyze and implement the missing functionality? It is astonishing at what pace this project has become relatively stable so I am sure there will be a way to also fix that somehow.

 

@PurpShell I remember you went a lot through IndexedDB for wwebjs when md has launched. Have you see any kind of stack for the to-be-resynced messages or similar?

adiwajshing:
The issue is likely in the way you're storing your keys, if you're using SQL or some other DB where there is a delay in storing/retreiving data -- there is a chance you use an incorrect signal session to encrypt/decrypt a message -- causing the issue you're seeing. 

This is very likely not an issue with the library

jackcokk:
There indeed is a delay. Do the auth state updates get ACK'd by Baileys? Could there be a way to make them wait for an event (db notifies successful save)?

I am asking because what may be possible is that Baileys during reconnecting loads an older set of auth whilst the current one is still being saved to the db. This would mean that Baileys has ACK'd the new set but launched the session with an incomplete set.

[...]

PR #696, opened 2021-09-22
Multi Device Support

adiwajshing:
This is still very early. Won't be merged anytime soon.

PR #890, opened 2021-11-18
Fixed Cannot read property 'key' of undefined

ilyaborodin:
Messages can be empty. I added logger and check this.

Снимок экрана 2021-11-18 в 23 29 07

So I added a condition to avoid the error

#874


PurpShell (approved):

adiwajshing:
messages should never be empty, this is solving the symptom not the actual cause of the problem.
Also I think I resolved this issue in a previous commit.

canove:
@adiwajshing, found a consistent way to reproduce this. 

You need to start two connections at the same time. If connection ONE sends a message to connection TWO, this error will throw.

Adding a `if !messages[0] return` immediately "resolves" it. But, as you pointed, we better find the cause.
```
 ev.on('messages.upsert', async ({ messages }) => {
        if (!messages[0]) {
          console.log("Messages array", messages)
          return;
        }
        var _a;
        const chat = { id: messages[0].key.remoteJid };
        const contactNameUpdates = {};
```

The messages array has no messages...
image

HUGEIT:
what you describe this may be an explanation for https://github.com/adiwajshing/Baileys/pull/833#issuecomment-968364048 that I ran into and maybe the cause or related is https://github.com/adiwajshing/Baileys/issues/870

adiwajshing:
Thanks. It's odd, I didn't know sending a stanza with 0 messages is possible -- I've added a patch accounting for it now:

https://github.com/adiwajshing/Baileys/pull/696/commits/27cf3954e8926313153d5be87b057770f6408842

HUGEIT:
> Thanks. It's odd, I didn't know sending a stanza with 0 messages is possible -- I've added a patch accounting for it now:

@adiwajshing have you had a case where this visibly happened + the log for it? if not, I know it sounds invasive but what do you think of anonymous telemetry for baileys? or would it be too hard to remove personal information for unhandled objects? Or just a few characters and then go up from there? just a few ideas..

Issue #868, opened 2021-11-15
[MD] Bad timestamp of messages after reconnecting


burstfreeze:
If I connect after having the bot offline for a while it will get all the messages. However the latest messages of the batch carry a wrong timestamp with them. Instead of the actual sent timestamp it shows the timestamp of when the connection was made.

Is this something that is expected to be in chats.update or somewhere? 

This is how the message looks, the bad time stamp is the second of connecting whereas the message was sent 4 hours earlier

```
{
  "messages": [
    {
      "key": {
 ..
      },
      "message": {
        "conversation": " "
      },
      "messageTimestamp": "1637000770",
      "status": "SERVER_ACK",
      "participant": " ",
      "pushName": " "
    }
  ],
  "type": "notify"
}
[
  {
    id: ' ',
    conversationTimestamp: 1637000770
  }
]
[
  {
    id: ' ',
    notify: ' '
  }
]
```

I am going through the logs right now. There are a few messages mentioning that message ID with other timestamps (cannot make sense of these) AND THEN a classic "No sender found .. failure in decrypting message".

It then did
```
Closing open session in favor of incoming prekey bundle
Closing session: SessionEntry {
```

and then 

```
{"level":30,"time": ,"pid": ,"hostname":" ","msgAttrs":{"from":" ","type":"text","id":" ","participant":" ","offline":"0","t":"1636984571","count":"6","notify":" "},"retryCount":1,"msg":"sent retry receipt"}
```

WHICH IS THE EXACT TIMESTAMP OF WHEN THE MESSAGE WAS SENT. Later we get

```
"messageStubType": "CIPHERTEXT",
      "messageStubParameters": [
        "No sender key for:
```

and eventually the actual message posted above. So basically we DO get the actual timestamp of the message once I guess. Now the question is whether it is expected behavior of the server to send the current timestamp instead of the sent timestamp OR if the client is expected to save the result from the message that carries the right timestamp in it.

hardianto-cpu:
just like me when I want to get the messagetype from the `message` property but the message property is not found and instead `messageStupType` appears

adiwajshing:
Cannot reproduce this, I'm using the timestamp sent to me by the server. I believe there is something else going on with your own code.

burstfreeze:
_Sorry for the long post but during writing a found a fix for this that we can add to Baileys_

[...]

PR #904, opened 2021-11-21
Emit event for true timestamp on retried messages as well instead of basically only emitting the current date.

burstfreeze:
https://github.com/adiwajshing/Baileys/issues/868


adiwajshing:
Can't you get the exact same amount of info from `messages.upsert`?

burstfreeze:
@adiwajshing this is what I have explained in the comments in the issue, nowhere does Baileys emit the real timestamp. You think someone would need the wrong timestamp, which is why I have added this event now so you can keep the bad one and everyone who needs the real one can listen to the new event.

adiwajshing:
Why wouldn't this work @burstfreeze ?

``` ts
const retriedMsgsTimestamps: { [_: string]: number } = { } 
ev.on('messages.upsert', ({ messages: [msg] }) => {
  if(msg.messageStubType === WAMessageStubType.CIPHERTEXT) }
    // store original timestamp
    retriedMsgsTimestamps[msg.key.id] = msg.timestamp
  } else {
    // restore original timestamp
    if(retriedMsgsTimestamps[msg.key.id]) msg.timestamp =  retriedMsgsTimestamps[msg.key.id]
  }  
})

```

burstfreeze:
Maybe I misunderstand something but in the `messages.upsert` I get output like this: 

`{"messages":[{"key":{"remoteJid":"123-123@g.us","fromMe":false,"id":"ABC","participant":"123@s.whatsapp.net"},"messageStubType":"CIPHERTEXT","messageStubParameters":["No sender key for: 123-123@g.us::123::0"]}],"type":"append"}`

which does not seem to include timestamps

adiwajshing:
That's my bad then @burstfreeze, this should resolve it:

https://github.com/adiwajshing/Baileys/pull/696/commits/d2e3abfe49088777dcd20256be7f282868957638

burstfreeze:
Thank you so much for your availability @adiwajshing this looks correct.

PR #912, opened 2021-11-22
Removed simple console.log

Manjit2003:


Manjit2003:
That console.log was spamming console

burstfreeze:
I do not think this should be removed from the example, knowing about fs access velocity is kind of useful.

adiwajshing:
I agree with @burstfreeze, it's fairly useful while testing to know when the auth state is being saved. If it's spamming your console, pls remove on your end only @Manjit2003

PR #913, opened 2021-11-22
[MD] Fixing unused variables and using the example.ts as a visible example

Dakauann:
Making the connection variable more explicit on the example code.


burstfreeze:
For the time being I actually prefer `sock` for MD and `conn` for legacy since it for now offers a more clear difference that is easier to spot. But in itself everything should be consistent

Dakauann:
> For the time being I actually prefer `sock` for MD and `conn` for legacy since it for now offers a more clear difference that is easier to spot. But in itself everything should be consistent

Normally on a real time coding i use this way to connect, and theres no problems, so i did this pull request to make It clean

adiwajshing:
So just copying the entire example script into the readme?

Dakauann:
> So just copying the entire example script into the readme?

Of course ;>

Dakauann:
> So just copying the entire example script into the readme?

I had just changed the socks in the example to conn, (the way I make the connection) but then I remembered the example, and found it more explanatory to use, it's up to you.

PR #915, opened 2021-11-23
[MD] Changing "sock" declaration to "conn" to avoid "sock is not defined"

Dakauann:
…ndefined declaration".

just changed the "sock" declaration to "conn" on events and on reconnect mode, If u test the old code you'll receive a "sock is not defined".


azudindaem:
Nothing to do with changing the name will solve the problem. Its just declaration. You can change on your self and no need to pr

Dakauann:
> Nothing to do with changing the name will solve the problem. Its just declaration. You can change on your self and no need to pr

it is meaningless to put something undeclared in an example.

README.md
```diff
@@ -44,30 +44,34 @@ TODO
 import makeWASocket from '@adiwajshing/baileys-md'
 
 async function connectToWhatsApp () {
+    // making a connection
     const conn = makeWASocket({
```
azudindaem, line 48: just change this to
const sock = makeWASocket({

azudindaem (changes requested):

azudindaem:
[Working example](https://github.com/adiwajshing/Baileys/tree/multi-device/Example)

Dakauann:
> [Working example](https://github.com/adiwajshing/Baileys/tree/multi-device/Example)

conn is more explanatory and most used by adiwajshing

azudindaem:
I believe the reason why adi uses sock on MD to differentiate with the legacy version. And the conn example in Readme is typo. I personally prefer sock for MD

PR #946, opened 2021-12-01
[MD] Added DeviceType

viniciusgdr:


adiwajshing:
What's the use case for this? This would be super unreliable, as it's very easy for any user to use a different kind of message ID -- spoofing the device type

viniciusgdr:
> What's the use case for this? This would be super unreliable, as it's very easy for any user to use a different kind of message ID -- spoofing the device type

The use case would be to identify the device that the user would be using, which someone could use for clients. Hypothetically, my client is using the web, so I will adapt it to use it on the web

As for Message ID spoofing, that would supposedly be rare. Because common WhatsApp users would not have this knowledge

adiwajshing:
The ID spoofing being rare argument is fair, however since the use case is personal to you and your clients & since it involves modifying the proto in an unfaithful way -- I would suggest this implementation stay on your end only. Or you could provide a util function to return the "expected" device type given the message ID

viniciusgdr:
> O argumento de falsificação de ID sendo raro é justo, no entanto, uma vez que o caso de uso é pessoal para você e seus clientes e uma vez que envolve a modificação do proto de uma forma infiel - eu sugeriria que esta implementação ficasse apenas não seu lado. Ou você pode fornecer uma função util para retornar o tipo de dispositivo "esperado" dado o ID da mensagem

I had cited a hypothetical situation, but you're right about modifying the proto

about creating a function, did you refer to "sock.getDevice()" for example?

adiwajshing:
@viniciusgdr doesn't even need to be a function of "sock" -- it can just be a pure function that returns the expected device type given a message ID

src/Utils/messages.ts
```diff
@@ -403,4 +403,12 @@ export const extractMessageContent = (content: WAMessageContent | undefined | nu
...
 	}
 	return content
-}
\ No newline at end of file
+}
+
+/**
+ * Returns the device predicted by message ID
+ */
+export const getDevice = (msg: WAMessage) => {
```
adiwajshing, line 411: this is the only function required. Also the input only requires the ID or the key at most, not the whole message.

The rest of the changes are unnecessary

PR #956, opened 2021-12-02
Update socket

mateuscdor:
let uploadInfo = await refreshMediaConn(false);
to
Let uploadInfo = await refreshMediaConn(true);

in lib/Socket/message-send.js: 318


Ethernit:
> let uploadInfo = await refreshMediaConn(false); to Let uploadInfo = await refreshMediaConn(true);
> 
> in lib/Socket/message-send.js: 318

Yes please update

PR #959, opened 2021-12-03
potential fix for QR scan. the server may expect these messages to be sent on reconnect after scan

HUGEIT:
discussion: should we have a way to kind of "ignore" the response? also maybe we do not need `await` here since we do not really care about the values.


adiwajshing:
I've made changes to parse & log when the queries are made -- we want to at least capture the results so they don't go unhandled.

 Also, since this is an additional setup step -- have moved it out of the base socket layer & into the chats layer.

PR #966, opened 2021-12-04
fix waiting for this message

LightningNeko:


burstfreeze:
How does renaming the variable help? Or do I misunderstand something about the spread operator 😅

HUGEIT:
@burstfreeze 

`devices` was defined before, adding itself to itself
https://en.wikipedia.org/wiki/Variable_shadowing

PR #971, opened 2021-12-05
fix group announce and restrict always false

LightningNeko:


adiwajshing:
Made a few more changes to make the messages in line with the original WA messages

PR #986, opened 2021-12-07
Add statusFind event.

AlanMartines:
I suggested adding this event to make monitoring the lib's built-in status easier.


adiwajshing:
I don't see the point -- "connection.update" already gives all this information?

burstfreeze:
5 things to mention here:

- your code has indents that are out of place, making it very hard to read for others
- your commits have also indented code that is not related to your changes - this should be avoided since they add nothing to the logic
- your comments have a lot of "back and forth". Usually you'd want to either re-fork with a clean commit history about every change or to squash the commits down to one commit
- you should follow the naming conventions of the maintainer (lower case, separated by `.`) instead of `'camelCase'`
- have you checked the `'connection.update'` event? It should contain most of the info you need. If it does not, maybe extend that event instead of creating a new one which may be somewhat redundant

AlanMartines:
> I don't see the point -- "connection.update" already gives all this information?

I'm monitoring the **"connection.update"** event, and most of the time it isn't called, where I can't catch the event when I disconnect the whats session by cell phone.

Bruno-Garcia-1 (approved):
ok

adiwajshing:
If the "connection.update" event isn't called -- don't you think it's a better idea to fix it than to build an entirely different event that does exactly the same thing ? @AlanMartines

mateuscdor:
> connection.update

As there are no events at all points
can be changed statusfind to conn.up..
and those who have conn.up can remove the statusfind and add the obj

adiwajshing:
That's not true at all @mateuscdor -- the event has all the connection related info one needs. Also, it still doesn't answer why connection.update cannot be modified to accommodate the same updates

mateuscdor:
Updates can also be done via "connection.update"
it was a very good suggestion.

AlanMartines:
I'm going to do some testing so by adding the conn.up event... at other points, I'll close this pull if there's this city I open again...

AlanMartines:
Tested but unsuccessful... reopening the pull.

adiwajshing:
Sorry mate, this PR won't be merged unless you can argue conclusively for the necessity of this new event & explain why connection.update cannot be "repaired" to serve the purpose

AlanMartines:
Ok...

PR #992, opened 2021-12-09
[MD] Make log "saving auth state" optional

guiguicdd:
This will make the console cleaner in large groups. Where this function run multiple times in order to check each member pre-key


mateuscdor:
just change the logger parameter: pin

é só alterar o parâmetro do logger: pino

guiguicdd:
const sock = makeWASocket({ printQRInTerminal: true, auth: state })

That is how my makeWASocket is. Do I need to insta pino?

Assim que está o makeWASocket. Devo instalar o pino?

mateuscdor:
me chama no whats  73 9 9177-6616

guiguicdd (commented):
sets logs as true by default

PR #1002, opened 2021-12-10
[MD] refactor: change pttAudio param to ptt in send audio message as a voi…

guiireal:
I'm sending audio as a Voice note using pttAudio param, but this not work for me. I change the pttAudio param to ptt and this works fine.


guiireal:
The test

image

The result

image

Issue #994, opened 2021-12-09
✅ [MD] - sendReadReceipt does not work with mobile disconnected from wi-fi ✅


mateuscdor:
**Version used:** Baileys-md (Multi-device)

the **.sendReadReceipt()** function does not work
with the mobile device disconnected from the internet.

The **.sendReadReceipt** function only works when the
mobile device is connected to internet.

Perform the test with the mobile device disconnected
from Internet.

```
client.ev.on('messages.upsert', async m => 
{ 

 const msg = m.messages[0]

  if(!msg.key.fromMe && m.type === 'notify') 
   {	
	const id          = m.messages[0].key.remoteJid
	const messageID   = m.messages[0].key.id 
	const participant = m.messages[0].key.remoteJid 
	
	await client.sendReadReceipt(id, participant, [messageID])
   }

}
```

![image](https://user-images.githubusercontent.com/90919904/145355207-2de8d30d-3555-431d-872e-83f82ce89b28.png)

![image](https://user-images.githubusercontent.com/90919904/145355434-7fe01f42-889b-4c16-9487-90b525c66ce1.png)

adiwajshing:
Mate you're supplying "remoteJid" as participant -- hence the error

mateuscdor:
Adiw I've tried in several ways with the device disconnected from the wifi doesn't work for me.

I tried in several ways, but with no success with the device disconnected.

barkerpedro:
Try:
```node
client.ev.on('messages.upsert', async m =>
{

 const msg = m.messages[0]

  if(!msg.key.fromMe && m.type === 'notify')
   {
	const id = m.messages[0].key.remoteJid
	const participant = client.user.id
	const messageID = m.messages[0].key.id
	
	await client.sendReadReceipt(id, participant, [messageID])
   }

}
```

mateuscdor:
@barkerpedro  me chama no whats 73 9 9177-6616 tenho um grupo só com br

mateuscdor:
@barkerpedro It also does not work with mobile device disconnected from the Wi-Fi

Também não funciona

mateuscdor:
@barkerpedro 
Resolved
https://github.com/adiwajshing/Baileys/pull/1016

PR #1016, opened 2021-12-14
fix sendReadReceipt()

barkerpedro:
I don't know if it's correct, but to fix the problem with the ".sendReadReceipt()" function when changing the status of a message received without internet on the cell phone.
As reported in #994

connected to internet:
Screenshot_4

disconnected on internet:
Screenshot_5


mateuscdor:
Fixed the problem, thanks for this pull
@barkerpedro

burstfreeze:
While this may work, it looks like it's overriding the client's privacy preferences since you somewhat invalidate the whole if statement.

From what you are writing it looks like this is not a matter of A or B depending on the setting but rather an A/B +C or even +C+D when offline, basically introducing one or more states depending on offline status additionally to the privacy status.

adiwajshing:
I agree with @burstfreeze why aren't privacy settings being respected?

barkerpedro:
> I agree with @burstfreeze why aren't privacy settings being respected?

Test it out by changing your privacy settings.

viniciusgdr:
Even changing it wouldn't make sense. Since you're disrespecting your privacy settings

Also, you made a case, which would be unnecessary as it would return everything the same. Like @burstfreeze mentioned

viniciusgdr:
> Even changing it wouldn't make sense. Since you're disrespecting your privacy settings
> 
> Also, you made a case, which would be unnecessary as it would return everything the same. Like @burstfreeze mentioned

I use this feature and it's working properly.

barkerpedro:
> > Even changing it wouldn't make sense. Since you're disrespecting your privacy settings
> > Also, you made a case, which would be unnecessary as it would return everything the same. Like @burstfreeze mentioned
> 
> I use this feature to read chats and it is working correctly.

But are you reading chats when your cell phone has no internet connection?

barkerpedro:
It's not really respecting the privacy settings. What would be the solution?

viniciusgdr:
> > > Even changing it wouldn't make sense. Since you're disrespecting your privacy settings
> > > Also, you made a case, which would be unnecessary as it would return everything the same. Like @burstfreeze mentioned
> > 
> > 
> > I use this feature to read chats and it is working correctly.
> 
> But are you reading chats when your cell phone has no internet connection?

no.

From what I've seen here, it's just reversing the order of privacy.

frmdeveloper:
What privacy do you mean?
I don't understand 😁😁

viniciusgdr:
This Correction has already been made by @lyfe00011 . The sendReadReceipt has probably already been resolved

barkerpedro:
No change required, fixed issue with correcting privacy settings in #1037

PR #1034, opened 2021-12-16
fix updateBlockStatus,fetchPrivacySettings

lyfe00011:


src/Socket/chats.ts
```diff
@@ -138,6 +138,8 @@ export const makeChatsSocket = (config: SocketConfig) => {
         await query({
             tag: 'iq',
             attrs: {
+		id: generateMessageTag(),
```
adiwajshing, line 141: don't need ID

lyfe00011: Oh it was there so

w3nder:
`src/Socket/messages-send.ts:43:60 - error TS2345: Argument of type 'string | number | BinaryNode' is not assignable to parameter of type 'BinaryNode'.
  Type 'string' is not assignable to type 'BinaryNode'.

43             privacySettings = reduceBinaryNodeToDictionary(content[0], 'category')`

When running the tsc command

PR #1038, opened 2021-12-16
Fix: Update BinaryNode  | privacySettings

mateuscdor:
will be fixed in compilation
https://github.com/adiwajshing/Baileys/blob/multi-device/src/Socket/messages-send.ts#L43


adiwajshing:
What exactly does this fix?

mateuscdor:
The tsc problem you fixed otherwise.
this was a second solution.
Thanks.

PR #1043, opened 2021-12-17
Fix #1010 by asking for full list of sync states when needed

### LINKED ISSUE #1010: [BUG] failed to verify LTHash -> no contact names in contacts.upsert, false group names in chats.update for individual chats
I have 2 instances with problems regarding contact names, I could gather a log for this. Most group names work just fine. All the contact names in `chats.set` are empty as expected, however on these sessions we do not get any `contacts.upsert` events at all.

What stands out in the log is

`"collections":["critical_block","critical_unblock_low","regular_high","regular_low","regular"],"error":"Error: failed to verify LTHash at 3 of critical_block\n > chat-utils.ts:409:23`

After that it does

```
"msg":"entering transaction"}
"msg":"resyncing critical_block from v0"}
"msg":"resyncing critical_unblock_low from v0"}
"msg":"resyncing regular_high from v0"}
"msg":"resyncing regular_low from v0"}
"msg":"resyncing regular from v0"}
```

After we get the list of pushnames ("notify") it does another weird thing:

```
groups.update
[{"id":"GROUP-TIMESTAMP@g.us","subject":"SUBJECT"}]

chats.update
[{"id":"CONTACT@s.whatsapp.net","conversationTimestamp":1636706051,"unreadCount":2404,"name":"SUBJECT"}]
```

(it applies the group name to an individual contact for no real reason)

There are a few more random

```
"resyncing main app state"}
"entering transaction"}
```

with some more `resyncing ABC from v123` but I don't know what information you need @adiwajshing

burstfreeze:
In #1010 I've described a problem with missing data occurring on some instances. It's due to the `has_more_patches` from the server that indicates you have to query again.

This PR checks for this flag and adds the result to the list.

To test this, add the following to `resyncAppState()`:

```
        var testList = []

        for (let i of result) {
            if(i.index.length > 1 && i.index[1] == 'regular_high') {
                testList.push(i)
            }
         }

        console.log(testList.length)
        
        console.log("results:",result.length)
        return result
    }
```

Running this gave me a count of 45 and 47 (2 runs). If I uncomment the `if(hasMorePatches) {` part, I get a length of 17 and 17 - in other words, it seems to work.

What is not taking into account here is the behavior with different `fromScratch` and `returnSnapshot` parameters. I cannot tell how or whether they can affect the outcome since I frankly do not know what they mean really 😅


burstfreeze:
For reference, this is how Tulir does it https://github.com/tulir/whatsmeow/blob/main/appstate.go#L51 

We could probably also do it like this but due to the different structure in Baileys the recursive way seems less invasive to the existing Baileys code.

src/Utils/chat-utils.ts
```diff
@@ -283,6 +283,8 @@ export const extractSyncdPatches = async(result: BinaryNode) => {
 
                 const syncds: proto.ISyncdPatch[] = []
                 const name = collectionNode.attrs.name as WAPatchName
+
+                const hasMorePatches = collectionNode.attrs.has_more_patches == 'true' ? true : false
```
adiwajshing, line 287: Can just replace with `collectionNode.attrs.has_more_patches == 'true'`

src/Socket/chats.ts
```diff
@@ -177,7 +177,7 @@ export const makeChatsSocket = (config: SocketConfig) => {
         })
     }
 
-    const resyncAppStateInternal = async(collections: WAPatchName[], fromScratch: boolean = false, returnSnapshot: boolean = false) => {
+    const resyncAppStateInternal = async(collections: WAPatchName[], fromScratch: boolean = false, returnSnapshot: boolean = false, isInner: boolean = false) => {
```
adiwajshing, line 180: I would strongly recommend against recursion -- might be better to put this in a while loop and then do it

burstfreeze:
I changed the logic to iterative now. Not sure if I broke something along the way but at least on my end the numbers add up:

```
RESULT MUTATIONS: 455
RESULT MUTATIONS: 41
TOTAL MUTATIONS: 496
```

image

PR #1066, opened 2021-12-24
feat: get latest version from whatsapp

gandol:
get latest version from whatsapp endpoint and implement it on exaple.ts

### DIFF
--- a/Example/example.ts
+++ b/Example/example.ts
@@ -9,12 +9,17 @@ import {
     ReconnectMode,
     ProxyAgent,
     waChatKey,
+    getLatestVersion,
 } from '../src/WAConnection'
 import * as fs from 'fs'
 
 async function example() {
+    const latestVersion = await getLatestVersion() //get latest version from whatsapp endpoint
     const conn = new WAConnection() // instantiate
     conn.autoReconnect = ReconnectMode.onConnectionLost // only automatically reconnect when the connection breaks
+    if(latestVersion){ // set to latest version
+        conn.version = latestVersion
+    }
     conn.logger.level = 'debug' // set to 'debug' to see what kind of stuff you can implement
     // attempt to reconnect at most 10 times in a row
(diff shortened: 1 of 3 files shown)


viniciusgdr:
It's probably not a good idea as there might be some code change in WhatsApp Web and some errors might occur.

viniciusgdr (commented):
It would not need to add a new dependency, it could be done with the "got" normally

HUGEIT:
the idea is cool but on a protocol change this will fuck up everything. I vote against merging this

gandol:
> It's probably not a good idea as there might be some code change in WhatsApp Web and some errors might occur.

it just optional I just create the function and implementation for this function is an example file, based on my current project i can't log in because just an old version

burstfreeze:
> it just optional

Except in your code it is not.

It is too risky and could result in unexpected behavior like feedback loops that could unnecessarily spam the server with requests which is exactly what we don't want to do since we want to play fair here. We are only guests on their system so we have to behave like guests and don't put excessive load on their service.

gandol:
> > it just optional
> 
> Except in your code it is not.
> 
> It is too risky and could result in unexpected behavior like feedback loops that could unnecessarily spam the server with requests which is exactly what we don't want to do since we want to play fair here. We are only guests on their system so we have to behave like guests and don't put excessive load on their service.

its true if the code for the checking inside loop or inside incomming message listener, but if the code just run once on try to login i think its a normal and the whatsapp web is using this method
image

burstfreeze:
No this is not what it's about. Let's say version 3.0.1 has a different handshake message and will close the connection on Baileys' attempt to connect, it will turn into an indefinite loop.

gandol:
> No this is not what it's about. Let's say version 3.0.1 has a different handshake message and will close the connection on Baileys' attempt to connect, it will turn into an indefinite loop.

hmmm i seee, if the code was merged and used on production will break all

PR #1070, opened 2021-12-24
fix listMessage

Pokoke-01:


README.md
```diff
@@ -309,6 +309,7 @@ const listMessage = {
   footer: "nice footer, link: https://google.com",
   title: "Amazing boldfaced list title",
   buttonText: "Required, text on the button to vie the list",
+  listType: 1,
```
PurpShell, line 312:
```suggestion
```
remove this, explained below

src/Types/Message.ts
```diff
@@ -67,6 +67,7 @@ type Listable = {
 
     /** Text of the bnutton on the list (required) */
     buttonText?: string
+    listType?: proto.ListMessage.ListMessageListType
```
PurpShell, line 70:
```suggestion
```
same over here

src/Utils/messages.ts
```diff
@@ -322,7 +322,8 @@ export const generateWAMessageContent = async(
 			buttonText: message.buttonText,
 			title: message.title,
 			footerText: message.footer,
-			description: message.text
+			description: message.text,
+			listType: message.listType
```
PurpShell, line 326:
```suggestion
			listType: 1
```

to keep everything simple, just make it default to 1. Now there is no need to modify existing code

PurpShell (changes requested):
changes, other than that should be sailing

S0c5:
Well, I liked more the previous changes, allow us to have more control over the messages also I don't know if it's a good idea to simplify the objects to build the Messages, given that those abstractions are not well documented and the only reference documentation is the proto code generated, we need to navigate in the code to know where exactly is assembled the message and how is the mapping.

PurpShell:
> Well, I liked more the previous changes, allow us to have more control over the messages also I don't know if it's a good idea to simplify the objects to build the Messages, given that those abstractions are not well documented and the only reference documentation is the proto code generated, we need to navigate in the code to know where exactly is assembled the message and how is the mapping.

We don't even know other types yet, adi explained it was getting too complex on the readme

this just makes it simple to the point where I can literally hook it up to my API without any processing. (pass in the content as is)


As for docs, look at type `AnyMessageContent`, I'd say clone baileys-md (`git clone https://github.com/adiwajshing/Baileys.git -b multi-device`), then generate the docs (`look at npm run`). 

You will be able to find the docs simple

PurpShell:
@adiwajshing thoughts?

PurpShell (approved):
lgtm

src/Utils/messages.ts
```diff
@@ -322,7 +322,8 @@ export const generateWAMessageContent = async(
 			buttonText: message.buttonText,
 			title: message.title,
 			footerText: message.footer,
-			description: message.text
+			description: message.text,
+			listType: 1
```
adiwajshing, line 326: Shouldn't this use the number from the ListType enum, instead of hardcoding it like this?

src/Utils/messages.ts
```diff
@@ -322,7 +322,8 @@ export const generateWAMessageContent = async(
 			buttonText: message.buttonText,
 			title: message.title,
 			footerText: message.footer,
-			description: message.text
+			description: message.text,
+			listType: proto.ListMessage.ListMessageListType[1]
```
adiwajshing, line 326: Bruh "SINGLE_SELECT" is the name of the property

PR #1074, opened 2021-12-25
Fix ContactsArrayMessage and add getBusinessProfile

BochilGaming:


PurpShell (approved):
lgtm

src/Types/index.ts
```diff
@@ -67,4 +67,17 @@ export type WABusinessProfile = {
...
+    description: string,
+    website?: string,
+    email?: string,
+    category: string,
+    business_hours: {
+        timezone: string,
+        business_config: WABusinessHoursConfig[]
+    }
+}
+
```
adiwajshing, line 82: would be better to use the same structure for legacy & MD

src/Socket/chats.ts
```diff
@@ -158,6 +158,49 @@ export const makeChatsSocket = (config: SocketConfig) => {
             ]
         })
     }
+    const getBusinessProfile = async (jid: string): Promise<WABusinessProfileMD | {}> => {
```
adiwajshing, line 161: Should just return undefined -- not an empty object

BochilGaming:
ok done

adiwajshing:
@BochilGaming please revert all formatting you've done -- stick to only changing code that needs to be changed

BochilGaming:
> @BochilGaming please revert all formatting you've done -- stick to only changing code that needs to be changed

sorry for that

PR #1089, opened 2021-12-27
Fix Played Status (message update event)

rafaelwedbiz:


src/Socket/messages-recv.ts
```diff
@@ -8,6 +8,11 @@ import { makeChatsSocket } from "./chats"
...
 
 const getStatusFromReceiptType = (type: string | undefined) => {
+
+    if(type === 'played') {
+        return proto.WebMessageInfo.WebMessageInfoStatus.PLAYED;
+    }
+
     if(type === 'read' || type === 'read-self') {
         return proto.WebMessageInfo.WebMessageInfoStatus.READ
     }
```
PurpShell, line 18: As @topkek on discord explained, this is better:
```suggestion
    if(type === 'read' || type === 'read-self') {
        return proto.WebMessageInfo.WebMessageInfoStatus.READ
    } else if(type === 'played') {
        return proto.WebMessageInfo.WebMessageInfoStatus.PLAYED;
    }
```

(there is always only 1 type of ack :shrug: )

PurpShell (changes requested):
:+1:, see my requested changes

PR #1144, opened 2022-01-13
change groupAcceptLink and add groupAcceptInvite

ArugaZ:
- change groupAcceptInvite to groupAcceptLink (code from https://chat.whatsapp.com/code)
- add groupAcceptInvite (code from groupInviteMessage)


src/Socket/groups.ts
```diff
@@ -129,11 +129,15 @@ export const makeGroupsSocket = (config: SocketConfig) => {
 			const inviteNode = getBinaryNodeChild(result, 'invite')
 			return inviteNode.attrs.code
 		},
-		groupAcceptInvite: async (code: string) => {
+		groupAcceptLink: async (code: string) => {
```
adiwajshing, line 132: don't see the need to change the name of the function.

src/Socket/groups.ts
```diff
@@ -129,11 +129,15 @@ export const makeGroupsSocket = (config: SocketConfig) => {
...
 			const results = await groupQuery('@g.us', 'set', [{ tag: 'invite', attrs: { code } }])
 			const result = getBinaryNodeChild(results, 'group')
 			return result.attrs.jid
 		},
+		groupAcceptInvite: async (jid: string, inviter: string, code: string, expiration: string) => {
```
adiwajshing, line 137: Might be more effective to use the invitemessage interface instead of passing so many arguments?

ArugaZ: yea i think so

PR #1164, opened 2022-01-18
Process unhandled connection error when changing proxy

ripreal:
If you set  a wrong proxy using Agent property in makeWASocket function then unhandled error might occur. The error leads to app crashing and you cannot catch it.


src/Socket/socket.ts
```diff
@@ -131,6 +131,8 @@ export const makeSocket = ({
                 },
             )
             return result as any
+        } catch (err) {
+            logger.error(`Connection closed. Reason: ${err}`)
```
adiwajshing, line 135: Logging the error doesn't do much mate, if the connection itself has failed -- you'll end up with a weird state, where nothing is coming in/going out

ripreal: I agree. I've found workaround that allows avoiding app exiting in case some error happens inside baileys 
```
process
  .on('unhandledRejection', (reason, p) => {
    // handling code
  })
  .on('uncaughtException', err => {
     // handling code
  });
```

PR #1191, opened 2022-01-27
chore: Update readme Discord link

jackcokk:
The old link leads to the now archived `#baileys` (non-MD) channel. This commit leads users to the `#read-this-first` channel that is probably the better starting point for newcomers.


PurpShell (approved):
supporting

adiwajshing:
Might be a good idea to put both

jackcokk:
> Might be a good idea to put both

The channel that is linked is for Baileys-legacy and it's read-only so it will be of little use.

We could also hot link the new Baileys channel but then people miss the FAQ that @PurpShell has created to solve 90% of the daily issues such as "can't send buttonMessage" etc.

PR #1245, opened 2022-02-16
check json format before saving

pe77:
sometimes the application corrupts the json and saves the corrupted data in the file.

Errors like '{' missing from end of file. Or ',' missing before "preKeys" key

This issue happened to me twice in 24 hours.


adiwajshing:
This removes some repetition I left in the code, which is great -- but if you're using a JSON file for your credentials in a production env -- you're bound to face a lot of issues

diezeldev:
Hi I had a similar issue just now with an empty JSON. I will fix today and submit

PR #1281, opened 2022-02-23
include peerDependenciesMeta

kaatt:
peer deps are required to be installed, optional deps aren't:

https://docs.npmjs.com/cli/v8/configuring-npm/package-json:

> peerDependencies: In some cases, you want to express the compatibility of your package with a host tool or library, while not necessarily doing a require of this host. This is usually referred to as a plugin. Notably, your module may be exposing a specific interface, expected and specified by the host documentation.

> optionalDependencies: If a dependency can be used, but you would like npm to proceed if it cannot be found or fails to install, then you may put it in the optionalDependencies object. This is a map of package name to version or url, just like the dependencies object. The difference is that build failures do not cause installation to fail. Running npm install --no-optional will prevent these dependencies from being installed.

we can also use `peerDependenciesMeta` to mark the peer deps as optional: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependenciesmeta


adiwajshing:
It used to be optionalDependencies, but due to a bug in yarn -- the deps were installed anyway. Hence, to prevent them from being installed automatically I made them peerDependencies, though technically not the right place to put them -- yarn no longer automatically installed the dependencies

kaatt:
updated to use peerDependenciesMeta

PR #1344, opened 2022-03-10
Add explicit types to socket factory functions

spkirby:
Add explicit types to socket factory functions. This prevents TypeScript error TS2742 being raised when Baileys is a dependency of another project.


diezeldev:
Hi, Baileys can be imported without issues here. There should be no need for what you are proposing here I think

PR #1352, opened 2022-03-11
Make disconnect reason in example more apparent when logged out.

diezeldev:
Hi this 'connection closed' message may confuse some. The only reason in that case is `loggedOut` so I think we should make this more obvious.


prototypeninja:
Hello. I recently saw your whatsapp api on github and I find it really great. I am ready to congratulate you for the work done and maintained. However, I have a small problem. I would like to create products in a whatsapp business catalog or a product catalog but I can't. I looked at the documentation but couldn't. I also wrote in discord but nothing. If you as a creator can help me I will be really overwhelmed. As I said above I am ready to buy you a coffee

vasani-arpit:
@prototypeninja I think you need to modify the library to do that. or you can go with chromium based library some of which allows that operation. They are heavy on resource though.

prototypeninja:
@vasani-arpit Can you help me to do it at this library? I like it's functionality. I'm ready to pay for it

diezeldev:
@prototypeninja hi please open an issue or use the discussion section for that. You are commenting on a merged PR here.

PR #1355, opened 2022-03-13
Add join group by groupInviteMessage

Schwartzblat:


README.md
```diff
@@ -732,6 +732,13 @@ Of course, replace ``` xyz ``` with an actual ID.
     ```
     Of course, replace ``` xxx ``` with invitation code.
 
+
+- To join the group using groupInviteMessage
+    ``` ts
+    const response = await sock.groupAcceptInviteV4(groupInviteMessage")
```
azudindaem, line 738: typo 
`const response = await sock.groupAcceptInviteV4("groupInviteMessage")`

azudindaem (changes requested):

README.md
Schwartzblat: Actually it supposed to be without "" at all :(
Fixed.

azudindaem:
there is no need to add new functionality to an existing one

ferhacks:
.

thisfardi (approved):

PR #1379, opened 2022-03-18
fix off-by-one error in downloadContentFromMessage

kaatt:


adiwajshing:
Closing, because expecting `endByte` to be exclusive like `array.slice()`

PR #1382, opened 2022-03-19
Added ReactionMessage for React User Message with Emoji

amisama:
image

### DIFF
--- a/src/Utils/messages.ts
+++ b/src/Utils/messages.ts
@@ -263,8 +263,7 @@ export const generateWAMessageContent = async(
 		}
 
 		m.extendedTextMessage = extContent
-	}
-	else if('contacts' in message) {
+	} else if('contacts' in message) {
 		const contactLen = message.contacts.contacts.length
 		if(!contactLen) {
 			throw new Boom('require atleast 1 contact', { statusCode: 400 })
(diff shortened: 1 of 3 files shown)


src/Utils/messages.ts
```diff
@@ -276,7 +277,11 @@ export const generateWAMessageContent = async(
 		}
 	} else if('location' in message) {
 		m.locationMessage = WAProto.LocationMessage.fromObject(message.location)
-	} else if('delete' in message) {
+	}
```
burstfreeze, line 280: Please try to stay within the coding style of the repo.

Check the lint scripts in package.json to verify or fix your style so it matches the rest.

amisama: sorry I'll fix it

adiwajshing:
Nice. What can be added:
1. firing `messages.update` with the reaction when a reaction message is sent/recv

amisama:
> Nice. What can be added:
> 
> 1. firing `messages.update` with the reaction when a reaction message is sent/recv



> Nice. What can be added:
> 
> 1. firing `messages.update` with the reaction when a reaction message is sent/recv
Its there on message.upsert, should I put it in messages.update too ?

IMG-20220321-WA0005

adiwajshing:
@amisama I meant call update for the original message that received the reaction

Issue #1387, opened 2022-03-20
[BUG] Timeout 408 error when setting ephemeral messages


sudomaaz:
**Describe the bug**
When trying to set message timer on , the function does not terminate and break with timeout. 

**To Reproduce**
```await sock.groupToggleEphemeral(group_id, 7 * 24 * 60 * 60);```
``` await sock.sendMessage(jid, { disappearingMessagesInChat: WA_DEFAULT_EPHEMERAL })```

Both these return timeout 408 error. **It only works when turning off message timer by passing 0 or false.**

Can validate by using these functions.

sudomaaz:
I have opened a PR for same. After making the necessary change in my local now its working fine. 

https://github.com/adiwajshing/Baileys/pull/1388

sudomaaz:
This PR is merged so closing the issue.

PR #1388, opened 2022-03-20
ephemeral key is invalid. Valid key is expiration

sudomaaz:


sudomaaz:
in the function groupToggleEphemeral since to pass ephemeral information the key "ephemeral" is used which is invalid and results in 408 timeout. The valid key to pass is "expiration". tested and verified on local.

sudomaaz:
Issue reported due to this 

https://github.com/adiwajshing/Baileys/issues/1387

HUGEIT:
can you check whether this still works in legacy? just making sure we don't break things

sudomaaz:
> can you check whether this still works in legacy? just making sure we don't break things

Yes key ephemeral still works in legacy its only changed to expiration in MD

HUGEIT:
just realized there is a different src folder for legacySocket 
in that case it's probably a very nice pr, I think it will be merged

PR #1393, opened 2022-03-21
chore: update proto

kaatt:


adiwajshing:
@kaatt using an alphabetically sorted Protobuf for easier diffs in the future now:

https://github.com/adiwajshing/Baileys/commit/51a397c27c9c8d38b7ca71413de68c0a5263ee72

PR #1404, opened 2022-03-23
Fix empty file written in saveState

diezeldev:
Hi when you close the process during writing it will leave the auth_info_multi.json file empty. The write is not atomic so it can happen.

Now we first write to a temp file and once that's done we replace the real file with the temp file. 

This means that if the process closes we still have our original auth_info_multi.json instead of losing all credentials. It may be in an older state but at least it's not trashing the whole session entirely, we can still reconnect with an outdated json.

Reference:

https://stackoverflow.com/a/39060492/17502680
https://stackoverflow.com/a/26262644

### DIFF
--- a/src/Utils/auth-utils.ts
+++ b/src/Utils/auth-utils.ts
@@ -115,18 +115,22 @@ export const initAuthCreds = (): AuthenticationCreds => {
 /** stores the full authentication state in a single JSON file */
 export const useSingleFileAuthState = (filename: string, logger?: Logger): { state: AuthenticationState, saveState: () => void } => {
 	// require fs here so that in case "fs" is not available -- the app does not crash
-	const { readFileSync, writeFileSync, existsSync } = require('fs')
+	const { readFileSync, writeFileSync, existsSync, renameSync } = require('fs')
 	let creds: AuthenticationCreds
 	let keys: any = { }
 
 	// save the authentication state to a file
 	const saveState = () => {
 		logger && logger.trace('saving auth state')
+
+		let temporaryFilename = `${filename}.temp`
+
 		writeFileSync(
-			filename,
+			temporaryFilename,
 			// BufferJSON replacer utility saves buffers nicely
 			JSON.stringify({ creds, keys }, BufferJSON.replacer, 2)
 		)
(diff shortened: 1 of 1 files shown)


Georgiobarcelona (commented):
En qué consiste la fusión

diezeldev:
Hi @Georgiobarcelona this fusion is about `auth_info_multi.json` sometimes being `0 bytes` due to a write error. He fixes the problem.

HanzCEO:
I don't really think this should be merged.

Because, this could be implemented by users themselves.

Especially when leaving `.temp` dust file on success or failing operations.

diezeldev:
Hi @HanzCEO you confuse mv with cp. mv does not create a duplicate, that’s what cp does.

HanzCEO:
> Hi @HanzCEO you confuse mv with cp. mv does not create a duplicate, that’s what cp does.

Thanks for the clarification. But then, when your patch is interrupted by kill signal it will still leaves dust `.temp` file.

Why not listening to SIGINT? Instead of band-aiding like this?

diezeldev:
I think being able to interrupt between the write and the mv is impossible. Obviously not impossible in theory but since both files reside in the same directory on the same drive it is as instantaneous as it can be. 

Even if you were to deliberately try to interrupt the process at that point you wouldn’t be able to do that by external intervention. And then, should you be able to somehow manage to do that the file will be gone on next update, potentially seconds later so you wouldn’t even notice. 


To sum it up, it’s a play between the somewhat likely case of having your session removed compared to the unlikely case of having a duplicate file on your disk for 3 seconds until baileys restarts. I would rather have the inconvenience of a file being there for 3 seconds without me even noticing than to re-scan the new session. 


That being said I would to see your SIGINT approach, I do not know what that is.

diezeldev:
Do not stale please

paulocorcino:
So, I believe that my problems occurred when the docker instance rebooted (due memory over loads, normally I wait for this) , and the same time that Baileys updating the credentials. I created a workaround, now I do a credential backup after each load success.
But is normal now the credentials grown up during time or my approach is wrong.

diezeldev:
Do not stale please

PR #1436, opened 2022-03-30
New example added

Alhar6i:
The example extends the multi-device example by allowing users to make connections and manage multi WhatsApp numbers on the same app.


gbenmelej:
Hello. I am trying to use and got this error

image
```

Phone PhoneNo2 is using WA v2.2208.14, isLatest: true
PhoneNo2 connection update [object Object]
Phone PhoneNo1 is using WA v2.2208.14, isLatest: true
PhoneNo1 connection update [object Object]
(node:13408) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'decode' of undefined
    at Object.processHandshake (C:\Users\german\Documents\BaileysMD\Baileys\src\Utils\noise-handler.ts:119:63)
    at WebSocket.validateConnection (C:\Users\german\Documents\BaileysMD\Baileys\src\Socket\socket.ts:178:24)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

adiwajshing:
I don't think adding a map & calling the example function a few times can really be called "multi sessions". Don't believe this PR is adding value to the project in its current state

Alhar6i:
> I don't think adding a map & calling the example function a few times can really be called "multi sessions". Don't believe this PR is adding value to the project in its current state

You are right, It was a wrong description of the function provided.
However, I just tried to share my experience with people who might be interested in that.
I respect your decision and believe that you might see something we don't.

Fonnte:
any drawback using this method @Alhar6i ?

Alhar6i:
> any drawback using this method @Alhar6i ?

@Fonnte I haven't made any investigation yet.  in the next two week I might be able to do an experimental test using at least 20 account.
I'll try to write my feedback here.

gbenmelej:
Today I try to use this method with 6 lines
With 2 lines it has worked perfectly, thanks!

Alhar6i:
> Today I try to use this method with 6 lines With 2 lines it has worked perfectly, thanks!

Thank you, good to hear that.

PR #1439, opened 2022-03-31
chore: Bump macOS version

burstfreeze:
The current version in Baileys is about 2 years old. We should bump this before they lock out old OS versions.


kaatt:
The version is extracted from `navigator.userAgent`, which is still `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.60 Safari/537.36` (10.15.7)

Chrome has decided not to update the User-Agent. So Baileys behavior would be incorrect after merging this.

burstfreeze:
@kaatt you are correct, thank you for sharing. The background is pretty interesting https://chromestatus.com/feature/5452592194781184, apparently websites aren't prepared for a prefix of `11` so a hacky workaround is to have the increments capped.

PR #1444, opened 2022-04-01
fix: No message present in quoted media on downloadMediaMessage

viniciusgdr:


HanzCEO:
Hello, your changes break backward compatibility by changing the function argument type.
Also, i don't think your changes fixes the problem.

viniciusgdr:
> Hello, your changes break backward compatibility by changing the function argument type. Also, i don't think your changes fixes the problem.

Really, but I don't know if you noticed, but when you want to download the mention of the message instead of the body of the message, it returns the error "No Message Present". In this instead of passing the "WAMessage" passes the "WAMessageContent" directly

HanzCEO:
With your solution, this doesn't work and create incompatibility
image

"No message present" occurred when any of the if statements in [this function](https://github.com/adiwajshing/Baileys/blob/5aa64f2c3967a89d22ae41c9db0c654e9a394da7/src/Utils/messages.ts#L485) is not matched. Perhaps, if possible, you can add one if condition to process regular message (not template message as implemented there)

viniciusgdr:
> With your solution, this doesn't work and create incompatibility image
> 
> "No message present" occurred when any of the if statements in [this function](https://github.com/adiwajshing/Baileys/blob/5aa64f2c3967a89d22ae41c9db0c654e9a394da7/src/Utils/messages.ts#L485) is not matched. Perhaps, if possible, you can add one if condition to process regular message (not template message as implemented there)

Understandable, I'll be closing PR provisionally and fixing it soon.

PR #1505, opened 2022-04-18
Fix null reference in remoteJid for ack READ

jeferssonlemes:
I noticed that when it's an ack read, the lib always returns the remoteJid as null, I debugged it for a long time in order to understand and I saw that the message arrives with the following structure:
```
{
  tag: "receipt",
  attrs: {
    from: "55xxxxxxxxxx:20@s.whatsapp.net",
    type: "read",
    id: "BAE54B1A91F891A7",
    participant: "55yyyyyyyyyyy@s.whatsapp.net",
    t: "1650305681",
  },
  content: null,
}
```

I don't know if this is the right fix, but I also know that I can't change it there in the **areJidsSameUser**.
Anyway, to simulate it, just test when it returns the read status of the message.


jeferssonlemes:
I noticed that after I merged into the last version, some things are weird regarding the ack, I'll investigate.

PR #1539, opened 2022-04-26
fix: `undefined` on unpin action

burstfreeze:
If you pin a chat, the event gives us a timestamp.

If you unpin a chat while Baileys is running, the `captureEventStream` will just say

`"chats.update","data":[{"id":"12345@s.whatsapp.net"}]}` without any information on the pin which is bad 😅

And the `chats.update` isn't helpful either because it just says `undefined` for the pin. This PR makes the event look like the following in the `captureEventStream`:

`"chats.update","data":[{"id":"12345@s.whatsapp.net","pin":false}]}`

so we can reliably obtain `pin` updates again.


adiwajshing:
So, the "undefined" was intentional -- if you want to print it out -- "null" may be a better option

PR #1548, opened 2022-04-27
Preview link fix

ilyaborodin:
matched-text - should be equal to link in text. So need to use url from regex in this field, because  link-preview-js can return url in another format and preview will not work in this case. 
Examples: 
text: "https://google.com" - link-preview-js will return "https://www.google.com/" and preview will not work
text:  "https://www.google.com" - link-preview-js will return "https://www.google.com/" and preview will work

Also link-preview-js does not support links without http/https prefixes. So if matched url does not have this prefixes, need to add them, before send text to link-preview-js library.

I am not good at regex, so just need to change regex to support all this variants: google.com, www.google.com, https://google.com. Current regex does not catch http/https prefixes.


ilyaborodin:
https://github.com/ilyaborodin/Baileys/commit/424a8c7eb000ac8407e9685d21c6eac0c6ac6ab1

I created this temp fix for me in order not to go into regex.

IMG_5995

src/Utils/link-preview.ts
```diff
@@ -19,15 +19,19 @@ export type URLGenerationOptions = {
  * Given a piece of text, checks for any URL present, generates link preview for the same and returns it
  * Return undefined if the fetch failed or no URL was found
  * @param text the text containing URL
+ * @param firstMatchedUrl first matched URL in text
  * @returns the URL info required to generate link preview
  */
 export const getUrlInfo = async(
```
adiwajshing, line 25: This helps quite a bit, but can you constrict the PR to only the `link-preview` file -- I don't believe you need changes in 5 files, only need to match the URL using the regex here in this util and set the matched text. 

The other alternative is to match the URL in the regex inside the `messages` util and pass only the url to the link preview generation function, this would be more efficient since it only requires one regex run

src/Socket/messages-send.ts
```diff
@@ -485,8 +485,8 @@ export const makeMessagesSocket = (config: SocketConfig) => {
 						logger,
 						userJid,
 						// multi-device does not have this yet
-						getUrlInfo: text => getUrlInfo(
-							text,
+						getUrlInfo: firstMatchedUrl => getUrlInfo(
```
adiwajshing, line 488: don't need this change

src/Types/Message.ts
```diff
@@ -165,7 +165,7 @@ export type MediaGenerationOptions = {
     mediaUploadTimeoutMs?: number
 }
 export type MessageContentGenerationOptions = MediaGenerationOptions & {
-	getUrlInfo?: (text: string) => Promise<WAUrlInfo>
+	getUrlInfo?: (firstMatchedUrl: string) => Promise<WAUrlInfo>
```
adiwajshing, line 168: redundant change

src/Utils/link-preview.ts
```diff
@@ -18,17 +18,16 @@ export type URLGenerationOptions = {
 /**
  * Given a piece of text, checks for any URL present, generates link preview for the same and returns it
  * Return undefined if the fetch failed or no URL was found
- * @param text the text containing URL
+ * @param firstMatchedUrl first matched URL in text
```
adiwajshing, line 21: no need to change the variable name, just the doc is fine

src/Utils/link-preview.ts
```diff
@@ -18,17 +18,16 @@ export type URLGenerationOptions = {
...
 export const getUrlInfo = async(
-	text: string,
+	firstMatchedUrl: string,
 	opts: URLGenerationOptions = { thumbnailWidth: THUMBNAIL_WIDTH_PX, timeoutMs: 3000 }
 ): Promise<WAUrlInfo | undefined> => {
 	try {
 		const { getLinkPreview } = await import('link-preview-js')
-
-		const info = await getLinkPreview(text, { timeout: opts.timeoutMs })
+		const info = await getLinkPreview('https://' + firstMatchedUrl, { timeout: opts.timeoutMs })
```
adiwajshing, line 30: What if firstMatchedUrl already has an https or http prefix?

ilyaborodin: Your regex doest catch http/https in links
Снимок экрана 2022-04-29 в 15 33 38

adiwajshing: Regardless, may be best to add a check to just future proof the code

ilyaborodin: Added check

PR #1553, opened 2022-04-29
Delete auth file on logout in examples

sreza78:
There is lots of issue i see in this community about connecting and authenticating the whatsapp session.

One of the problems that confused me alot is about when log out happens from whatsapp app, the auth file should be deleted to create the new one of it.
So i fixed it in the examples for others to pay attention to that


sreza78:
Please review this RP

SL-Alpha-X:
Use `fs.unlink()`

sreza78:
:(

PR #1562, opened 2022-05-03
feat: Returns an object with information about the invite code's group

LeonardoBein:
This PR is intended to add the functionality of get group information by invite code.

Example: 
```ts
 const response = await sock.groupGetInviteInfo("xxx");
 console.log("group information: " + response)
```

Reponse `GroupMetadata` interface:

```ts
interface GroupMetadata {
    id: string
    owner: string | undefined
    subject: string
    /** group subject owner */
    subjectOwner?: string
    /** group subject modification date */
    subjectTime?: number
    creation: number
    desc?: string
    descOwner?: string
    descId?: string
    /** is set when the group only allows admins to change group settings */
    restrict?: boolean
    /** is set when the group only allows admins to write messages */
    announce?: boolean
    /** number of group participants */
    size?: number
    // Baileys modified array
    participants: GroupParticipant[]
    ephemeralDuration?: number
}
```


LeonardoBein:
Please, review this PR.

HanzCEO:
I actually like this. But you should make the object more readable.
What even is `s_t` and `s_o`, for example.

src/Socket/groups.ts
```diff
@@ -137,6 +137,23 @@ export const makeGroupsSocket = (config: SocketConfig) => {
 			const result = getBinaryNodeChild(results, 'group')
 			return result.attrs.jid
 		},
+		groupGetInviteInfo: async(code: string) => {
+			const results = await groupQuery('@g.us', 'get', [{ tag: 'invite', attrs: { code } }])
```
adiwajshing, line 141: I think you can just use `extractGroupMetadata` for this instead of writing another mapping function?

LeonardoBein: It really is possible, thank you very much for the help.

ferhacks:
.

adiwajshing:
@LeonardoBein can you resolve the conflict please -- can merge after that

LeonardoBein:
> @LeonardoBein can you resolve the conflict please -- can merge after that

done!

PR #1611, opened 2022-05-12
update profile pic in legacy version

pe77:
now it works,

I had to do it because I still use the legacy version in some cases


pe77:
up

PR #1655, opened 2022-05-23
Group Invite Message Added

ravindu01manoj:


adiwajshing:
This is a good idea but I feel can be executed better, eg.
``` ts
// utility function to fetch invite details for a group & generate the invite message
const groupInvite = await sock.generateGroupInviteMessage('abcd@g.us')
await sock.sendMessage('1234@s.whatsapp.net', { groupInvite })
```

ravindu01manoj:
```ts
   // utility function to fetch invite details for a group & generate the invite message
   const groupInvite = await sock.generateGroupInviteMessage('abcd@g.us','1234@s.whatsapp.net')
   await sock.sendMessage('1234@s.whatsapp.net', { groupInvite })
   ```

File: src/Socket/groups.ts [hunk 1]
```diff
@@ -122,6 +122,24 @@ export const makeGroupsSocket = (config: SocketConfig) => {
 				]
 			)
 		},
+		generateGroupInviteMessage: async(jid: string, participant: string) => {
+			const result = await groupQuery(
```
adiwajshing, line 126: Isn't this just the add participant function, where you're using the failure to fetch the code?

ravindu01manoj:
```ts
  if(groupInvite)
```
users can be use if method to catch failure... or can you  suggest something ?

adiwajshing: I don't see why `groupInviteCode` can't be used?

ravindu01manoj: groupInviteCode and add fail group invitation codes are not same... if we use groupInviteCode... not loading group info for receiver (tested)

PR #1685, opened 2022-06-01
feat: send text message defalt type conversation

huangzhong0406:


adiwajshing:
Not sure why this PR is required

huangzhong0406:
> Not sure why this PR is required

When I need to send a message of type conversation ，I can't set it, it's always extendedTextMessage，so I made a modification

adiwajshing:
I think you should just modify your code instead of updating the default type like this

huangzhong0406:
ok , thanks

Issue #1693, opened 2022-06-04
[BUG] the sent retry receipt issue, an observation.


Alhar6i:
**Describe the bug**
As you can see the console yells:

{"level":30,"time":"2022-06-04T07:13:19.060Z","pid":17508,"hostname":"DESKTOP-S2KD8VD","msgAttrs":{"from":"9665038*****@s.whatsapp.net","type":"text","id":"3A0F6277819C7B2FAB7E","notify":"┘å╪º┘è┘ü ╪º┘ä╪¡╪▒╪¿┘è","t":"1654326799"},"retryCount":3,"msg":"sent retry receipt"}

{"level":30,"time":"2022-06-04T07:13:16.403Z","pid":17508,"hostname":"DESKTOP-S2KD8VD","msgAttrs":{"from":"9665038*****@s.whatsapp.net","type":"text","id":"3A0F6277819C7B2FAB7E","notify":"┘å╪º┘è┘ü ╪º┘ä╪¡╪▒╪¿┘è","t":"1654326754"},"retryCount":2,"msg":"sent retry receipt"}

{"level":30,"time":"2022-06-04T07:23:25.182Z","pid":17508,"hostname":"DESKTOP-S2KD8VD","msgAttrs":{"from":"status@broadcast","type":"media","id":"3ABF2A255E6DA785528B","participant":"9665034*****@s.whatsapp.net","notify":"╪¬┘à╪│┘â ╪¿┘à┘å ┘è╪¡╪¿┘â ┘ê┘è╪▒╪º┘â ╪¼┘å╪¬┘ç","t":"1654327404"},"retryCount":1,"msg":"sent retry receipt"}

some extra info:

{
  key: {
    remoteJid: 'status@broadcast',
    fromMe: false,
    id: '3AFEC86CDF7DBD04CF5E',
    participant: '9665556XXXXX@s.whatsapp.net'
  },
  messageTimestamp: 1654340015,
  pushName: 'XXXXX',
  messageStubType: 2,
  messageStubParameters: [
    "ENOENT: no such file or directory, open 'D:\\Business\\Projects\\WhatsApp\\WICServer\\auth_info_Controller\\sender-key-status@broadcast::9665556XXXXX::0.json'"
  ]
}

If you pay attention you will notice that all the messages are of "sent retry receipt"
[...]

Issue #1589, opened 2022-05-09
Closing session: SessionEntry Removing old closed session


phenriquec:
Hi,

I have a situation where this message appears after connect and doesn't disappear anymore, apparently in a loop!

does anyone know the reason?

09/05/2022 11:15:11 Closing session: SessionEntry {
  _chains: {
    BbHoTRLL0giFLDtD9egTL6VxQhCJBO13EfdjX8HOGTRk: { chainKey: [Object], chainType: 2, messageKeys: {} },
    'BZiCBY0E8LSXMvGdeeGT5hWLU5f3jVis35i/XPfWhTd2': { chainKey: [Object], chainType: 1, messageKeys: {} }
  },
  registrationId: 589168037,
  currentRatchet: {
    ephemeralKeyPair: {
      pubKey: <Buffer 05 98 82 05 8d 04 f0 b4 97 32 f1 9d 79 e1 93 e6 15 8b 53 97 f7 8d 58 ac df 98 bf 5c f7 d6 85 37 76>,
      privKey: <Buffer 98 20 32 b8 8e 18 a3 c4 87 b7 63 a4 5d e9 a9 c0 93 3b f5 50 cb 76 01 b9 dd a7 18 f6 6f dd ad 49>
    },
    lastRemoteEphemeralKey: <Buffer 05 b1 e8 4d 12 cb d2 08 85 2c 3b 43 f5 e8 13 2f a5 71 42 10 89 04 ed 77 11 f7 63 5f c1 ce 19 34 64>,
    previousCounter: 0,
    rootKey: <Buffer 52 57 6e 9f d0 a8 b5 3e 10 b0 29 cf dd 95 43 5a c8 50 8b 97 9f f1 24 50 1f 9f 19 7f 66 a5 cd d3>
  },
  indexInfo: {
    baseKey: <Buffer 05 74 18 94 c5 25 b3 4c 45 82 f7 48 2b 5d fb cc c4 1c 03 74 96 57 f6 b6 90 9b 23 e1 04 73 29 32 41>,
    baseKeyType: 2,
    closed: -1,
    used: 1652105710807,
    created: 1652105710807,
    remoteIdentityKey: <Buffer 05 b3 67 1b 3b df 22 9a f2 01 d1 1e f2 ef ff 43 f9 56 f4 57 36 13 da 34 d6 62 4d d9 68 94 1d 9b 50>
  }
}

09/05/2022 11:15:10 Removing old closed session: SessionEntry {
[...]

PR #1696, opened 2022-06-04
Fix file name error on user-multi-file-auth-state.ts

Alhar6i:
user-multi-file-auth-state fails to write/read the sender-key-status@broadcast*.json files because they have a colon.
To fix it, for now, I replaced the colon with a dash.
Hope you find a better solution.
This might solve the issues below
https://github.com/adiwajshing/Baileys/issues/1589#issue-1229804957
https://github.com/adiwajshing/Baileys/issues/1693#issue-1260676356


adiwajshing:
Before this, have you investigated why does a colon prevent writing a file?

Alhar6i:
> Before this, have you investigated why does a colon prevent writing a file?

Oh yeah, I forgot to say I'm a windows user! Windows do not allow it.
It might not the case on Linux and macOS
https://stackoverflow.com/a/13298479

adiwajshing:
Ah I see, I'm a Mac user -- so that's why I'm guessing I'm oblivious to this

PR #1700, opened 2022-06-05
fix: reaction message doesn't appear when send to a group

fadiinho:
This PR fixes when you send a reaction message to a group it doesn't appear, it's caused by the agent number in `participant` property.


src/Utils/process-message.ts
```diff
@@ -23,8 +23,9 @@ const MSG_MISSED_CALL_TYPES = new Set([
 
 /** Cleans a received message to further processing */
 export const cleanMessage = (message: proto.IWebMessageInfo, meId: string) => {
-	// ensure remoteJid doesn't have device or agent in it
+	// ensure remoteJid and participant doesn't have device or agent in it
 	message.key.remoteJid = jidNormalizedUser(message.key.remoteJid!)
+	message.key.participant = message.key.participant ? jidNormalizedUser(message.key.participant!) : null
```
adiwajshing, line 28: use undefined instead of null please

PR #1702, opened 2022-06-05
Add option to distinguish between join group types.

Schwartzblat:


azudindaem (approved):
thanks

adiwajshing:
I think you can just listen for `messages.upsert` and use the stub type from that to accomplish what you're trying to do. There isn't a need to modify the baileys source for that

Schwartzblat:
I can do it but it less comfortable and it makes more sense to add it to group-participants.update.
It's not changing anything only makes group-participants.update event more usable.

adiwajshing:
I don't think adding a "messageStubType" is a good idea for the "group-participants.update" event

Schwartzblat:
Why not? It gives more information on the event and explained what cause this change.

PR #1712, opened 2022-06-08
refactor(Events): Omitted `Partial<>` around `Contact`

HanzCEO:
`Partial<Contact>` is just the same as normal `Contact`.
Because, in every emit of `contacts.upsert`,
Baileys _always_ set the `Contact.id` no matter what.

Thus, adding `Partial` is redundant.


HanzCEO:
I think this is a bad idea.
Might confuse docs in the future

PR #1718, opened 2022-06-10
fix: Delete Message For Me

DwiRizqiH:


hassanfuad96:
This fix issue for iphone sender.

azudindaem (approved):
looks good

moskoweb:
Doesn't this function allow you to delete more than one message at the same time? I surrounded her with several messages, and she deleted only the first one.

PR #1725, opened 2022-06-11
fix #1724 change connectionLost status code to 504(Gateway timeout)

### LINKED ISSUE #1724: How to figure connection is lost or QR attempts ended.
If you check `connectionLost` and `timeout` both has same status code `408` so it's hard to figure out with the status that connection is lost or just qr ref attempts ended. Only way is to read the error message. But I think we should have to make it better by adding some kind of extra property to let know that connection was lost or qr attempts fail?

AxeemHaider:


azudindaem:
Not necessary.. we can see payload message for any reason..

AxeemHaider:
Reading message I think is not relatable way If in future a single character change from message you will lose. Why not to make a different statusCode ?

azudindaem:
yeah.. 408 is related to timeout and connection lost normally 502 or 504.. that make sense

AxeemHaider:
504 make sense in this case, That's why I changed and create PR, waiting to merge

viniciusgdr:
504 - Server Side
408 - Client Side

When the connection is lost it is on the client side, the server has nothing to do with it. So I don't support change.

adiwajshing:
Yeah, this won't be merged

AxeemHaider:
Have you checked this issue #1724

PR #1615, opened 2022-05-13
Add 'extras' to groupParticipantsUpdate

JonnyShuali:
The feature added a new object inside groupParticipantsUpdate object result, which contains the invite code and expiration timestamp while adding a new participant to a group.

PR #1727, opened 2022-06-11
feat: Add 'extras' to groupParticipantsUpdate

JonnyShuali:
The feature added a new object inside groupParticipantsUpdate object result, which contains the invite code and expiration timestamp while adding a new participant to a group.
All formatted  as asked in #1615


adiwajshing:
What's missing:
1. types for the extra property
2. linting of the files

PR #1783, opened 2022-06-22
Type from mediaKey in DownloadableMessage

moskoweb:


adiwajshing:
Don't see why this has been done?

PR #1799, opened 2022-06-24
Add option to distinguish between join group types.

Schwartzblat:


Schwartzblat:
The last one closed because stale.

PR #1837, opened 2022-06-29
Lost connection code correction

moskoweb:
The lost connection code, which can be defined as a banned connection as well, was with the same code as ```timedOut```, I adjusted the order of the codes, and define ```connectionLost``` as code 403, returned when the connection is lost.

### DIFF
--- a/src/Types/index.ts
+++ b/src/Types/index.ts
@@ -49,6 +49,7 @@ export enum DisconnectReason {
     connectionClosed = 428,
     connectionReplaced = 440,
     badSession = 500,
+    banned = 503,
     restartRequired = 515,
 }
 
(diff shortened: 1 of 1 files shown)


adiwajshing:
So `403` is *sometimes* banned as far as I know, `408` is connection lost/timed out -- these shouldn't be changed

moskoweb:
> So `403` is _sometimes_ banned as far as I know, `408` is connection lost/timed out -- these shouldn't be changed

I did the tests and banned 5 numbers just to test which status actually sends, and it is correct that when banning the first return is 503, which I added, then the connection loss occurs, issuing a 403 as status.

In my view, I don't think it makes sense to define two constants with the same value within the same Enum, if they can be emitted different.

That's why I added the banned field and changed the lost connection value.

adiwajshing:
That may be, but changing "connectionLost" to 403 will break all existing baileys implementations & you won't be able to distinguish between a regular connection break and being banned. 

Can add the `403` banned statusCode to DisconnectReason -- but changing the existing code won't be accepted

moskoweb:
Why was it not accepted?

PR #1845, opened 2022-06-30
fix: add `senderTimestampMs` automatically in reactMessage

zennn08:
# Remove reaction message
I have tried with below code to remove reactMessage, but it doesn't work.
``` ts
const reactionMessage = {
    react: {
        text: "",
        key: message.key
    }
}
const sendMsg = await sock.sendMessage(id, reactionMessage)
```
then I found a way by adding `senderTimestampMs: Date.now()` to make it work.


README.md
```diff
@@ -411,8 +411,9 @@ const sendMsg = await sock.sendMessage(id, listMessage)
 
 const reactionMessage = {
     react: {
-        text: "💖",
-        key: message.key
+        text: "💖", // use an empty string to remove the reaction
+        key: message.key,
+	senderTimestampMs: Date.now() // this is needed if you want to remove the reaction
```
adiwajshing, line 416: may be worth it to just include this automatically in `src/utils/messages.ts` instead of requiring users to input?

zennn08: ok i will change it

adiwajshing:
PR looks good, only need to lint

zennn08:
done sir

PR #1852, opened 2022-07-02
Update generics.ts

DzOfc:


adiwajshing:
Not changing the default ID generation mechanism

PR #1868, opened 2022-07-05
whatsapp web version

ravindu01manoj:


barkerpedro:
You can put the following link instead:
```typescript
const URL = 'https://web.whatsapp.com/check-update?version=1&platform=web'
```

ravindu01manoj:
> You can put the following link instead:
> 
> ```ts
> const URL = 'https://web.whatsapp.com/check-update?version=1&platform=web'
> ```

okey :)

adiwajshing:
Ah no, would like to have control over the version being specified by baileys. In the future, if there is a major update -- we don't want baileys to break just because we're fetching and running the latest version from WA Web.

I would recommend adding another function to fetch from WA Web, and keeping the original one as is

Sew01RaviduManoj01KingAndQueen:
> Ah no, would like to have control over the version being specified by baileys. In the future, if there is a major update -- we don't want baileys to break just because we're fetching and running the latest version from WA Web.
> 
> I would recommend adding another function to fetch from WA Web, and keeping the original one as is

.

src/Utils/generics.ts
```diff
@@ -254,6 +254,28 @@ export const fetchLatestBaileysVersion = async() => {
 	}
 }
 
+/**
+ * A utility that fetches the latest web version of whatsapp.
+ * Use to ensure your WA connection is always on the latest version
+ */
+export const fetchLatestWaWebVersion = async() => {
+	const URL = 'https://web.whatsapp.com/check-update?version=1&platform=web'
```
adiwajshing, line 262: should be `url` not screaming-case

ravindu01manoj: oh

PR #1871, opened 2022-07-05
fix: link preview use https protocol

LightningNeko:
link preview doesn't work if https:// is not used in matched-text


src/Utils/link-preview.ts
```diff
@@ -46,7 +46,7 @@ export const getUrlInfo = async(
 
 			return {
 				'canonical-url': info.url,
-				'matched-text': text,
+				'matched-text': previewLink,
```
adiwajshing, line 49: I don't think this would work -- since matched-text needs to be present inside the message text which it isn't now

PurpShell: > I don't think this would work -- since matched-text needs to be present inside the message text which it isn't now

i think this was just resolved

LightningNeko: > I don't think this would work -- since matched-text needs to be present inside the message text which it isn't now

whatsapp does not generate or show previews if they do not start with https or http

Issue #1425, opened 2022-03-28
[BUG] The promise rejected with the reason "1006".     


zakirkun:
I have problem 

> [ERROR] 21:31:37 UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "1006".

burstfreeze:
You have to share more of the log, this single line is not enough for us to fix something.

Share the whole log but make sure to delete personal identifiers such as phone numbers before people start calling you 😅

makanSukros:
Me too getting error like this

uncaughtException [UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "1006".] {
  code: 'ERR_UNHANDLED_REJECTION'
}

I've wrapped all async await functions with try catch but the above error still appears I don't know where it came from 😅

zakirkun:
> Me too getting error like this
> 
> uncaughtException [UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "1006".] { code: 'ERR_UNHANDLED_REJECTION' }
> 
> I've wrapped all async await functions with try catch but the above error still appears I don't know where it came from 😅

many socket connectin open i think

aislanspeedio:
Same problem, here!!!

PR #1881, opened 2022-07-08
hoist catch into mutex callback body, so promise rejection is handled gracefully

mwhitworth:
otherwise the internal promise rejection is unhandled


src/Socket/chats.ts
```diff
@@ -526,11 +526,10 @@ export const makeChatsSocket = (config: SocketConfig) => {
 
 		await (
 			processingMutex.mutex(
-				() => resyncAppState(ALL_WA_PATCH_NAMES, ctx)
-			)
-				.catch(err => (
```
adiwajshing, line 531: I'm fairly certain it gets handled outside too -- can you send the stack trace for this?

mwhitworth: Yes, you are right, as promises "auto-unwrap". I'll close this. I've noticed where the `.catch` is placed (either inside the mutex body or outside) is inconsistent between methods in general though.

I think there a few floating promises in the codebase that cause unhandled rejections like #1425. I'll use the [no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md#no-floating-promises) linter rule to audit this, and open separate PRs here (if I do find anything!)

Issue #1897, opened 2022-07-12
[BUG] `assertMediaContent` doesn't consider `imageMessage` inside other types (e.g. `buttonsMessage`)


ponei:
**Describe the bug**
[`assertMediaContent` ](https://github.com/adiwajshing/Baileys/blob/master/src/Utils/messages.ts#L696) is used when downloading media from messages as fail safe to messages that don't contain media.
It works fine for most simple types, but doesn't consider more complex messages like `buttonsMessage`
It >can< have `imageMessage` inside it, but it's not required. `headerType` would be `IMAGE` in this case.

More message types could have `imageMessage` inside them and assert would fail those too.

**To Reproduce**
1. Receive a `buttonsMessage` with `headerType` `IMAGE`
2. Try to download the image with `downloadMediaMessage`

**Expected behavior**
Buffer containing the media

Environment (please complete the following information):
Node version: 14.18.0
Network: Localhost
Baileys version: 4.2.0

ponei:
Fixed as of [this commit](https://github.com/ponei/Baileys/commit/24dd76771eabe11c2804052dd87dd5b0a834082e)
As time of writing this (13/07/2022) this is only on the bleeding edge Github branch and not 4.2.0

PR #1898, opened 2022-07-12
fix: check for more messages with `imageMessage` in `assertMediaContent`

ponei:
If a message with an image requires re-uploading, the flow goes as follow:
`updateMediaMessage` -> `assertMediaContent` -> decides if message contains media
This is fine for most messages, but doesn't check for all messages that can contain `imageMessage`
One of those is `buttonsMessage`. If `headerType` is `IMAGE` and the image requires re-uploading, `assertMediaContent` will throw early, blocking the re-upload, even though the message contains media.
This pull request adds more checks to `assertMediaContent` that includes: `buttonsMessage`, `templateMessage` and `interactiveMessage` 

Should close [this issue](https://github.com/adiwajshing/Baileys/issues/1897)


adiwajshing:
I think `extractMessageContent` already does what you're looking for?

ponei:
Oh yeah, my bad.
Was testing agains't `4.2.0` which still has the faulty `normalizeMessageContent` 😅

PR #1922, opened 2022-07-17
fix: detectLinks was not being used as described in docs

### LINKED ISSUE #1921: Missing type for detectLinks?
As said [here](https://github.com/adiwajshing/Baileys#notes), you can use `detectLinks: false` to not generate previews when sending a message with an url. I have searched in the code and I don't found anything about `detectLinks`.
```ts
...
   // will detect links & generate a link preview automatically (default true)
    detectLinks: true
...
```
image

fadiinho:
Closes #1921


adiwajshing:
Hmm, `detectLinks` shouldn't exist -- to mimic the behaviour you can use:
```
conn.sendMessage(jid, { text: 'test', linkPreview: null })
```

fadiinho:
So maybe we can remove it from the readme?

PR #1959, opened 2022-07-27
chore: update version

xxirfanx:
updated to latest version.


frmdeveloper:
Actually it's not important. Because automatic updates are available
```js
const { version, isLatest } = await fetchLatestBaileysVersion()
let sock = makeWASocket({
      auth: state,
      version,
      logger: P({ level: 'silent' })
      markOnlineOnConnect: false,
      linkPreviewImageThumbnailWidth: 10,
      syncFullHistory: true
    })
```

PR #1967, opened 2022-07-30
Do not duplicate button submissions

lopesclayton:
when sending a message with null buttons, the message is sent twice
1 - to the recipient
2 - to the sender


src/Socket/messages-send.ts
```diff
@@ -398,7 +400,9 @@ export const makeMessagesSocket = (config: SocketConfig) => {
 
 					if(!participant) {
 						devices.push({ user })
-						devices.push({ user: meUser })
+						if (hydratedButtons.length > 0) {
```
adiwajshing, line 403: don't see how this is preventing any duplicate button submissions?

lopesclayton: sorry, I expressed myself badly the code prevents the message from being sent in duplicate, ie if I send a message ckm template butttons,. it is received by the recipient and by the sender (the message is sent twice)

adiwajshing: no, this logic doesn't seem right to me, why are we pushing to our own device again? the steps below would ensure that anyway

PR #1981, opened 2022-08-03
chore: update version

xxirfanx:
**updated to latest version.**


adiwajshing:
No need to update so quickly, can just use `fetchLatestWaWebVersion` in your code

PR #2007, opened 2022-08-08
Fix baileys install through github

PurpShell:
There is an issue while installing Baileys through github (`yarn add github:adiwajshing/Baileys` or `yarn add git+https://github.com/adiwajshing/Baileys`), where the prepare script fails the typescript compile due to missing types in dependencies.. (mainly keyed-db and @types/ws)

this ensures all modules are installed, though using yarn or npm is up to debate..


PurpShell:
Tf happened to the linter

fadiinho:
This not works. The `prepare` script runs on `yarn install`, so if you make `prepare` execute `yarn install`, this will be an infinity loop. You can see it [here](https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts): `Runs on local npm install without any arguments`
image

PurpShell:
> 

hmm

fadiinho:
I guess I just discovered the solution.
I was reading the Baileys `package.json` and noticed that tsc was not a dev dependency.
As you can read [here](https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts):
`NOTE: If a package being installed through git contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.`
So yarn tries to run `tsc` but isn't installed. Maybe if add tsc as a dev dependency it will work.

PurpShell:
> I guess I just discovered the solution.
> I was reading the Baileys `package.json` and noticed that tsc was not a dev dependency.
> As you can read [here](https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts):
> `NOTE: If a package being installed through git contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.`
> So yarn tries to run `tsc` but isn't installed. Maybe if add tsc as a dev dependency it will work.

tsc means typescript compile, it is found in the "typescript" package

PR #2012, opened 2022-08-09
switch to discussion

frmdeveloper:
So that not all problems are poured into issues


PurpShell:
I prefer discord

adiwajshing:
Would be good to just add both, if possible

frmdeveloper:
Looks like something's wrong

PR #2018, opened 2022-08-10
docs: add how to pin and delete a chat

fadiinho:
This PR adds missing docs about how to pin/unpin and delete a chat.


fadiinho:
Unstale comment

PR #2027, opened 2022-08-11
Changing getDevice shapes

kauanzito:
med I bring an "explanation"nces between OS, and I bring an "explanation"


viniciusgdr:
has no reason to change the answers, it could break codes that depend on this function

kauanzito:
> has no reason to change the answers, it could break codes that depend on this function

well, it will be more explanatory that way, and not to mention that it will be easier to understand. A user might get confused by getting their deviceType as "web" (which is what is currently there), but with the term "WhatsApp Web", it will know what it is.

viniciusgdr:
but there is no reason to send the user which device he is connected to, because he already knows. But if you really want to send it to the user you can change the answers with the conditionals

kauanzito:
Ah yes. it is really not necessary for the user to see the type of device he uses, but it will be easier for him to understand (as I had already said), and also to not need to modify his own code (as you mentioned) every time Baileys makes a update...

adiwajshing:
there isn't a convincing reason to update this function, I'd recommend a wrapper around this in your own code.

PR #2048, opened 2022-08-16
[GitHub version] TS Compile error fixes and GH Pages revamp

PurpShell:
The socket documentation is a mess, so I tried my best at fixing it, also this should hopefully fix the typescript compile error:
```
src/Socket/messages-send.ts:12:14 - error TS2742: The inferred type of 'makeMessagesSocket' cannot be named without a reference to '@adiwajshing/baileys/node_modules/@types/ws'. This is likely not portable. A type annotation is necessary.
```
that occured for every socket instance

This changes the layout of the docs though, but exports all the types too, so it looks better and cleaner


PurpShell:
@adiwajshing

adiwajshing:
this is like way too much maintenance to do the types, I don't think the TS import issues are from baileys itself

PurpShell:
> this is like way too much maintenance to do the types, I don't think the TS import issues are from baileys itself

image

the package was not resolved because it failed to compile

PR #1990, opened 2022-08-04
List and template bug fix

maulalwi:
I tried to get rid of the uncommentment in the script and the script ran normally again

PR #2098, opened 2022-08-27
Update TemplateMessage (Fix)

ds6:


viniciusgdr:
This method does not work for iOS users, even crashing them. Therefore, there is still no definitive solution.

This is the clone of https://github.com/adiwajshing/Baileys/pull/1990

adiwajshing:
Please search through the repo before creating new PRs & issues 🙏

PR #2101, opened 2022-08-27
fix template message

xdlyy404:


viniciusgdr:
This method does not work for iOS users, even crashing them. Therefore, there is still no definitive solution.

This is the clone of #1990

Kuroosh:
LMAO xD

PR #2125, opened 2022-09-06
Store with Multi Files

moskoweb:
Using the normal single file function ends up bursting or causing slow recording of large amounts of conversations and messages, splitting to multiple files has worked better for storing information.


src/Store/make-in-memory-store.ts
```diff
@@ -387,6 +387,53 @@ export default (
 				const json = JSON.parse(jsonStr)
 				fromJSON(json)
 			}
+		},
+		writeToMultiFiles: (folder: string) => {
+			// require fs here so that in case "fs" is not available -- the app does not crash
+			const { existsSync, mkdirSync, writeFileSync } = require('fs')
+			if (folder.includes('.json')){
```
adiwajshing, line 394: isn't this redundant?

src/Store/make-in-memory-store.ts
```diff
@@ -387,6 +387,53 @@ export default (
...
+		},
+		readFromMultiFiles: (folder: string) => {
+			// require fs here so that in case "fs" is not available -- the app does not crash
+			const { readFileSync, existsSync } = require('fs');
+			if (folder.includes('.json')){
+				readFromFile(folder);
+			} else {
+				if (existsSync(folder)) {
+					logger.debug({ folder }, 'reading from file');
+					const chats = JSON.parse(
```
adiwajshing, line 420: if one of these reads fails -- won't it cause a crash?

src/Store/make-in-memory-store.ts
moskoweb: I don't see how it is, this validation is to make the migration from Single File to Multiple Files easier.

src/Store/make-in-memory-store.ts
moskoweb: I put verification of the files

moskoweb:
Why has it become obsolete?

amiruldev20:
this function can't be called when i try

moskoweb:
> this function can't be called when i try

try again

amiruldev20:
no work. error installed baileys

moskoweb:
> no work. error installed baileys

What error appears to you?

PR #2156, opened 2022-09-14
fix getImageProcessingLibrary

kaatt:


salykingleb:
Can you tell me how this commit helps?

PR #2170, opened 2022-09-19
Feat: updateGroupProfilePicture

guilhermeromer:
updateProfilePicture() was not working properly in groups and after some research I realized that the query to change specifically group profile pictures might have been changed.

Here I bring a small solution for that.


src/Socket/chats.ts
```diff
@@ -186,6 +186,27 @@ export const makeChatsSocket = (config: SocketConfig) => {
 		})
 	}
 
+	/** update the profile picture for a group */
+	const updateGroupProfilePicture = async(jid: string, content: WAMediaUpload) => {
```
adiwajshing, line 190: `updateProfilePicture` and `updateGroupProfilePicture` can likely very easily be combined into one function

guilhermeromer: You're right. I already made some changes

PR #2171, opened 2022-09-20
fix: groups.update

DwiRizqiH:
Fix groups.update not fire when desc groups changed


src/WABinary/types.ts
```diff
@@ -9,9 +9,9 @@ import * as constants from './constants'
 export type BinaryNode = {
     tag: string
     attrs: { [key: string]: string }
-	content?: BinaryNode[] | string | Uint8Array
+	content?: BinaryNode[] | string | Uint8Array | { content: string }
```
adiwajshing, line 12: What, why is this being modified when the actual decoding code isn't? And there's no need to add this either

PR #2190, opened 2022-09-24
Setting a default empty options AxiosRequestConfig<any>

Alhar6i:


Alhar6i:
for fetchLatestWaWebVersion

PR #2197, opened 2022-09-26
feat: add notification node type `picture`

ponei:
this pull request adds support for the notification node type `picture`
this notification is fired whenever someone changes the group's picture

when your group already has a picture and it's changed to a new one, the message would have the `photoChange` property
`photoChange` contains the thumbnail of the old and new picture
this property is only available when you first sync the messages; the notification doesn't have the necessary data to create it
since this isn't feasible (right now), no additional message processing is needed

additionally, this pull request also changes the chain of `else if`'s to a `switch` for more visibility

the notification looks like:
```json
    "frame": {
      "tag": "notification",
      "attrs": {
        "from": "group jid",
        "type": "picture",
        "id": "2797004126",
        "notify": "notify name",
        "t": "1664163305"
      },
      "content": [
        {
          "tag": "set",
          "attrs": {
            "author": "jid of who changed",
            "jid": "group jid",
            "id": "1664163304"
          }
        }
      ]
    }
```

the message looks like:
```json
      "participant": "jid of who changed",
      "messageStubType": "GROUP_CHANGE_ICON",
      "messageStubParameters": [
        "1664163304"
      ],
```

### DIFF
--- a/src/Socket/messages-recv.ts
+++ b/src/Socket/messages-recv.ts
@@ -194,7 +194,8 @@ export const makeMessagesRecvSocket = (config: SocketConfig) => {
 		const [child] = getAllBinaryNodeChildren(node)
 		const nodeType = node.attrs.type
 
-		if(nodeType === 'w:gp2') {
+		switch (nodeType) {
+		case 'w:gp2':
 			switch (child?.tag) {
 			case 'create':
 				const metadata = extractGroupMetadata(child)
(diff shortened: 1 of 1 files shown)


azudindaem (approved):

src/Socket/messages-recv.ts
```diff
@@ -430,7 +448,7 @@ export const makeMessagesRecvSocket = (config: SocketConfig) => {
 							id: node.attrs.id,
 							...(msg.key || {})
 						}
-						msg.participant = node.attrs.participant
+						msg.participant ??= node.attrs.participant
```
adiwajshing, line 451: didn't get the need for this change, otherwise PR looks good

azudindaem: I think there's a typo..

ponei: > didn't get the need for this change, otherwise PR looks good

[nullish assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment) means that it will set the participant to `attrs.participant` if the message doesn't already have a participant
since I already set the participant from the notification content (`attrs.author`) and the notification doesn't have the participant tag (`attrs.participant`), the old logic would override it to undefined

adiwajshing: Ah, I see what's happening

PR #2241, opened 2022-10-09
solution for message duplication with link preview

undefined-60:
It may not be the best solution but it works well.


undefined-60:
> This PR is stale because it has been open 6 days with no activity. Remove the stale label or comment or this will be closed in 2 days

await...

adiwajshing:
I'd recommend you do the de-duping via a store on your end

Issue #2161, opened 2022-09-15
[BUG] externalAdReply does not appear in text messages


SafwanGanz:
conn.sendMessage(m.chat, {text: 'hai', contextInfo: {"externalAdReply": {title: titlink, mediaType: 3, renderLargerThumbnail: false, showAdAttribution: true, body: deslink, thumbnail: image, sourceUrl: `${args[0]}`}}}, {quoted: m})

its not working

amiruldev20:
use  [this baileys](https://github.com/amiruldev20/Baileys)

IqbalzzX:
use relayMessage work

nadiadev7:
use baileys [https://github.com/amiruldev20/Baileys](Amirul Dev)  or [https://npmjs.com/package/baileys](this package) its work

SafwanGanz:
> use relayMessage work

Ok..

SafwanGanz:
> use [this baileys](https://github.com/amiruldev20/Baileys)

Its baileys bro

amiruldev20:
> > use [this baileys](https://github.com/amiruldev20/Baileys)
> 
> Its baileys bro

my baileys has been fixed about that bug

amiruldev20:
![Screenshot (3)](https://user-images.githubusercontent.com/73012169/190886730-dc3938d9-f5fa-43eb-bfaf-609a57576b97.png)

souravkl11:
> > > use [this baileys](https://github.com/amiruldev20/Baileys)
> > 
> > 
> > Its baileys bro
> 
> my baileys has been fixed about that bug

Why don't you make a fix pr on this repo? Make it!

Issue #1747, opened 2022-06-15
[BUG]


vynnshome:
contextInfo not working in type text.

azudindaem:
can you explain the context?

PR #2242, opened 2022-10-09
fix: extra context info not adding to messages

BlackAmda:
Can't add extra context info into messages.
Like, forwardingScore, isForwarded, externalAdReply etc...

Example: 
```js
await sock.sendMessage(id, { text: 'oh hello there', contextInfo: { forwardingScore: 1, isForwarded: true }})
// some random context info (can show a forwarded message)
```

Fix: #2161 #1747


BlackAmda:
closing for remove unwanted trailing spaces

PR #2243, opened 2022-10-09
fix: extra context info not adding to messages

BlackAmda:
Can't add extra context info into messages.
Like, forwardingScore, isForwarded, externalAdReply etc...

Example: 
```js
await sock.sendMessage(id, { text: 'oh hello there', contextInfo: { forwardingScore: 1, isForwarded: true }})
// some random context info (can show a forwarded message)
```

Fix: #2161  #1747


BlackAmda:
+1

BlackAmda:
+1

BlackAmda:
> This PR is stale because it has been open 6 days with no activity. Remove the stale label or comment or this will be closed in 2 days

+1

src/Utils/messages.ts
```diff
@@ -454,6 +454,11 @@ export const generateWAMessageContent = async(
 		m[messageType].contextInfo.mentionedJid = message.mentions
 	}
 
+	if('contextInfo' in message && message) {
```
adiwajshing, line 457: contextInfo could be undefined

src/Utils/messages.ts
```diff
@@ -454,6 +454,11 @@ export const generateWAMessageContent = async(
...
+		const [messageType] = Object.keys(m)
+		m[messageType].contextInfo = m[messageType] || { }
```
adiwajshing, line 459: merge this contextInfo into the existing one if it exists

src/Utils/messages.ts
```diff
@@ -454,6 +454,11 @@ export const generateWAMessageContent = async(
 		m[messageType].contextInfo.mentionedJid = message.mentions
 	}
 
+	if('contextInfo' in message && message) {
+		const [messageType] = Object.keys(m)
```
adiwajshing, line 458: messageType is defined before this, use it from there -- no need to define the same variable twice

PR #2251, opened 2022-10-11
Removing messages from msgRetryMap

Alhar6i:
The current logic deletes messages if the retry counter is greater than or equal to 5 which means that if the retry succeeded before reaching this threshold the message won't be deleted and the msgRetryMap will keep growing up.
Here we remove message ids from "msgRetryMap" if they have been successfully sent.


Alhar6i:
make the AxiosRequestConfig parameter optional in fetchLatestWaWebVersion

Alhar6i:
@adiwajshing any comments on this PR?

adiwajshing:
1. This doesn't account for the fact that different devices of the same user may not have received the message correctly -- only one device did
2. This clearly hasn't been tested, because msgRetryMap keys using `participant + id` not just `id`. So even via this PR, it won't do anything

Alhar6i:
Ok, many thanks

PR #2252, opened 2022-10-11
docs: How to forwading messages

Stazyu:


azudindaem (changes requested):
There is no need to change the existing guide

SheIITear:
The current is fine, this just makes it more confusing.

Issue #1622, opened 2022-05-15
Error when getting user profile pic item not found


AxeemHaider:
**Describe the bug**
I'm using edge version on Baileys, Everything was working and suddenly I'm getting this error this is because when I try to get profile pic of user

```
Error: item-not-found
    at Object.assertNodeErrorFree (/home/haider/codeplace/bitrespond/wa-backend/node_modules/.pnpm/github.com+adiwajshing+baileys@85a58fbc10caa92e726044336a7e6ff3f562fc62/node_modules/@adiwajshing/baileys/lib/WABinary/generic-utils.js:56:15)
    at query (/home/haider/codeplace/bitrespond/wa-backend/node_modules/.pnpm/github.com+adiwajshing+baileys@85a58fbc10caa92e726044336a7e6ff3f562fc62/node_modules/@adiwajshing/baileys/lib/Socket/socket.js:135:24)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at Object.profilePictureUrl (/home/haider/codeplace/bitrespond/wa-backend/node_modules/.pnpm/github.com+adiwajshing+baileys@85a58fbc10caa92e726044336a7e6ff3f562fc62/node_modules/@adiwajshing/baileys/lib/Socket/chats.js:310:24)
```

**To Reproduce**
1. Created a new connection

**Environment (please complete the following information):**
 - Is this on a server? YES
 - What do your `connectOptions` look like? 
 ```js
   makeWASocket({
      auth: creds,
    });
```
 - Do you have multiple clients on the same IP? NO
 - Are you using a proxy? NO

HUGEIT:
does the process crash or is it just the error?

AxeemHaider:
process crash.

I want out the way why this is happen if user does not have profile pic then this error come. I think this should be only error not to crash the process but this error crash the process

panalgin:
This also happens for me too, the process is crashed and there is no way to catch it via try/catch operation

`{"level":10,"time":"2022-10-19T15:35:05.418Z","pid":11896,"hostname":"DESKTOP-5GJQITL","msgId":"53049.63166-7","fromMe":false,"frame":{"tag":"iq","attrs":{"from":"***@s.whatsapp.net","type":"error","id":"53049.63166-7"},"content":[{"tag":"error","attrs":{"code":"404","text":"item-not-found"}}]},"msg":"communication"}
[...]

PR #2271, opened 2022-10-19
fix for item-not-found, when user does not have a profile picture set

panalgin:
Hi, regarding a stale issue #1622 

When the current user has no profile picture set, profilePictureUrl function triggers a "Boom" error upon receiving "item-not-found" response from the WA servers and this terminates the execution, since this Boom crash can not be caught or bypassed from the client code, I had to make this modification 

Because having a "null profile picture url" is not a fatal error, the **assetNodeErrorFree** function should bypass this error, instead of throwing a Boom-type crash.

Below is a simple fix, and I don't think it will affect anywhere on the project where "item-not-found" error should be regarded as a fatal crash.

Please feel free to comment and suggest fixes


adiwajshing:
Why not just:
```
sock.getProfilePicture(jid).catch(() => null)
```

The Boom is there so someone could distinguish between the errors, like 404s, 403s & 401s

panalgin:
> Why not just:
> 
> ```
> sock.getProfilePicture(jid).catch(() => null)
> ```
> 
> The Boom is there so someone could distinguish between the errors, like 404s, 403s & 401s

This doesn't work, the execution halts and node process gets killed immediately

panalgin:
Sorry it is my bad, seems like try catch block doesn't work but, using .catch directly worked, now it catches the boom. You can disregard this pull request, it's my poor TS skills 🤣

```TypeScript
              try {
                sock.profilePictureUrl(sock.user!.id, "image", 5000).then((url: string | undefined) => {
                  console.log("Pic: " + url);
                }).catch((error) => {
                  console.log("This gets caught");
                });
              }
              catch (error) {
                  console.log("This won't get caught");
              }

vienapps:
> Sorry it is my bad, seems like try catch block doesn't work but, using .catch directly worked, now it catches the boom. You can disregard this pull request, it's my poor TS skills 🤣
> 
> ```ts
>               try {
>                 sock.profilePictureUrl(sock.user!.id, "image", 5000).then((url: string | undefined) => {
>                   console.log("Pic: " + url);
>                 }).catch((error) => {
>                   console.log("This gets caught");
>                 });
>               }
>               catch (error) {
>                   console.log("This won't get caught");
>               }
> ```

How To convert To  Extension .js Sir....?

PR #2345, opened 2022-11-19
Fix connect example

juanmanavarro:
Added async to messages.upsert callback


adiwajshing:
this PR is for an outdated version, closing

PR #2362, opened 2022-11-24
Update README.md

naufkia:
line 500 variable must be templateMessage not buttonMessage


BlackAmda:
It's just a constant reference. Could be anything. 😁

Issue #2369, opened 2022-11-27
[BUG] Error: mutex start


zennn08:
**Describe the bug**
Spamming log error mutex start

`{"level":40,"time":"2022-11-27T06:47:18.034Z","pid":3132077,"hostname":"xxxx","stack":"Error: mutex start\n    at /root/xxx/bot/node_modules/@adiwajshing/baileys/lib/Utils/make-mutex.js:15:31\n    at Object.mutex (/root/xxx/bot/node_modules/@adiwajshing/baileys/lib/Utils/make-mutex.js:35:15)\n    at handleReceipt (/root/xxx/bot/node_modules/@adiwajshing/baileys/lib/Socket/messages-recv.js:348:29)\n    at execTask (/root/xxx/bot/node_modules/@adiwajshing/baileys/lib/Socket/messages-recv.js:549:20)\n    at processNodeWithBuffer (/root/xxx/bot/node_modules/@adiwajshing/baileys/lib/Socket/messages-recv.js:539:19)\n    at WebSocket.<anonymous> (/root/xxx/bot/node_modules/@adiwajshing/baileys/lib/Socket/messages-recv.js:569:9)\n    at WebSocket.emit (node:events:513:28)\n    at /root/xxx/bot/node_modules/@adiwajshing/baileys/lib/Socket/socket.js:222:35\n    at Object.decodeFrame (/root/xxx/bot/node_modules/@adiwajshing/baileys/lib/Utils/noise-handler.js:131:17)\n    at WebSocket.onMessageRecieved (/root/xxx/bot/node_modules/@adiwajshing/baileys/lib/Socket/socket.js:198:15)","waitOver":true,"msg":"possible mutex deadlock"}`

makanSukros:
+1

MaznAbdullah:
+1

azudindaem:
Is this occur after updating to the latest commit?

zennn08:
yes

Fonnte:
+1

azudindaem:
can you explain on house to reproduce this error.

erickythierry:
+1

but the bot does not stop responding
it just fills the logs with this error

zennn08:
> can you explain on how to reproduce this error.

You just need to update to the latest commit version and run the bot. after a few minutes, spam logs occur as above

zennn08:
[...]

PR #2453, opened 2023-01-05
Bugfix: "mutex start" warns are not on correct logger instance

panalgin:
Hello,

Regarding the issue at #2369 

Mutex start error is not logged through to the actual "pino" Logger instance, instead imports the pino Logger from scratch and overrides what the user preference of log warning levels, even when the `logger.level = "fatal"` is specificed, that mutex (logger.warn) is displayed on the terminal. This is pretty annoying when running multiple baileys instances in the same node process.

The solution is to use the "user-created logger instance", and pass it to the "makeMutex" function, which is how it is done for the rest of the app

Tested it locally and on my production server, which runs several hundreds of instances of baileys instances and those mutex warns are no longer bloating the logs.

Yet, why so many mutex start warns are happening, especially during rush hours of WhatsApp, is another question to be solved later

### DIFF
--- a/src/Socket/chats.ts
+++ b/src/Socket/chats.ts
@@ -34,7 +34,7 @@ export const makeChatsSocket = (config: SocketConfig) => {
 	let needToFlushWithAppStateSync = false
 	let pendingAppStateSync = false
 	/** this mutex ensures that the notifications (receipts, messages etc.) are processed in order */
-	const processingMutex = makeMutex()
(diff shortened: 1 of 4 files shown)


zennn08 (approved):

ahmedRSA (approved):

adiwajshing:
Thank you for bringing this up, I've removed that log now -- so am closing this PR

PR #2495, opened 2023-01-18
add support for labels on chats and messages

injitools:
open for discussion about realization

### DIFF
--- a/Example/example.ts
+++ b/Example/example.ts
@@ -121,8 +121,8 @@ const startSock = async() => {
 
 							if(store) {
 								const labels = Object.values(store.labels)
-								await sock.setChatLabel(msg.key.remoteJid!, labels[0].id)
-								await sock.setMessageLabel(msg.key.remoteJid!, msg.key.id!, labels[1].id)
+								await sock.addChatLabel(msg.key.remoteJid!, labels[0].id)
+								await sock.addMessageLabel(msg.key.remoteJid!, msg.key.id!, labels[1].id)
 							}
 						}
 					}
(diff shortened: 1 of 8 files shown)


azudindaem:
suggest to change variable **labaled** 

Would you mind to put this implementation in chatModify so user can use sock.chatModify

injitools:
> suggest to change variable **labaled**

"labaled" from proto description of protocol. i'm not change it. it's standart whatsapp parameter


> Would you mind to put this implementation in chatModify so user can use sock.chatModify

it's already in chatModify and i add two helper functions for creating mod for it.
```

const setChatLabel = async(jid: string, labelId: string, labaled: boolean = true) => {
	return chatModify({
		type: LabelAssociationType.JID,
		labelId,
		labelAssociationAction: { labaled },
	}, jid)
}
const setMessageLabel = async(jid: string, messageId: string, labelId: string, labaled: boolean = true) => {
	return chatModify({
		type: LabelAssociationType.MESSAGE,
		labelId,
		messageId,
		labelAssociationAction: { labaled },
	}, jid)
}
```
so you just call it from sock
```
sock.setChatLabel(msg.key.remoteJid!, labels[0].id)
sock.setMessageLabel(msg.key.remoteJid!, msg.key.id!, labels[1].id)
```

azudindaem:
you may use "labelled" instead of "labaled"

it would be nice if you can make it like

``` ts
  await sock.chatModify({
    addLabel: {
      label: '1',
    },
  },
  '123456@s.whatsapp.net')
  
  await sock.chatModify({
    removeLabel: {
      label: '1',
    },
  },
  '123456@s.whatsapp.net')
  ```

injitools:
> you may use "labelled" instead of "labaled"
> 
> it would be nice if you can make it like

make it work like you example

also add `addMessageLabel` and `removeMessageLabel`

Issue #2503, opened 2023-01-19
[BUG] I don't get new message event when reconnection is established


Guilherme-j10:
**Describe the bug**
event messages.upsert literally doesn't work when credentials are used to log back in

**To Reproduce**
1. Created a new connection
2. Closed & used saved credentials to log back in
3. Then try to receive a new message and you will not receive

**Expected behavior**
When doing a reconnection with the session files I expect to receive the messages normally

**Environment:**
 - Is this on a server?: No
 - Do you have multiple clients on the same IP?: Yes
 - Are you using a proxy?: No

PR #2504, opened 2023-01-19
fix: messages.upsert

Guilherme-j10:
As I had reported here #2503, I was having problems receiving messages through the messages.upsert event. When I connected the device for the first time, it worked normally, but when I connected for the second time through the authentication credentials, the event stopped working, so debugging the code in order to solve my problem, I saw that there was a return preventing the event from being triggered, so I removed it. I make it clear that I don't know if this has any impact in general in the context of the application but it worked for me


hansputera:
up

adiwajshing:
this PR would create duplicate events:
1. one that is released after the buffer is complete
2. one that is released immediately

I don't believe the issue exists presently on the master branch, and I cannot particularly reproduce your case.

PR #2520, opened 2023-01-26
feat: convert function structure to classes

SamuelScheit:
## Why is this needed?
The current functional approach makes it quite difficult to add support for the native mobile api, because some functions need to have different implementations.
Using a class-based approach would make it easier to extend the project to support both the web and native api while keeping the code organized, maintainable, and easy to understand.

## Implementation
I did not change the implementation logic of any function and tried to keep the changed lines as low as possible.
I've converted `makeSocket...()` into the following classes: `Socket, Chats, MessagesSend, MessagesReceive, Groups` that all extend each other.
I've converted the return variables of `makeSocket` into class properties and methods and added `this` where previously it used the return result of makeSocket.
I've checked the example and everything still works as expected.
Sometimes GitHub does not show a good diff for changed indentation, I recommend opening the PR in an IDE.


adiwajshing:
Could you please elaborate more on why it's hard to make changes, with an example, using the functional approach?

PR #2528, opened 2023-01-31
Set linkPreview with undefined

moskoweb:
In generateWAMessageContent, it is checked if it is undefined, and not if it is null.

https://github.com/adiwajshing/Baileys/blob/7923233829f6a5fd1a53059ae6f6eb2b6d222782/src/Utils/messages.ts#L288


moskoweb:
@adiwajshing

PR #2672, opened 2023-03-27
chore: update connecting example in the docs

sostenesapollo:
This PR fixes the example with the connection code (in the docs) that gives an error when you first uses it
image

I don't want to understand other things when I first use the lib, I just want to connect. 
I spent some hours to fix this, so please merge it. Thanks


sostenesapollo:
Easiest as just install the lib using `yarn add `@adiwajshing/baileys` , install hapi, and just run `npx ts-node index.ts` and don't give you any anoyng error

azudindaem (changes requested):
This PR is not necessary. The existing sample file in the example folder is complete enough for the first use

sostenesapollo:
@azudindaem can you try to copy and paste that and run, pls ?

sostenesapollo:
@azudindaem that's because you've just cloned the repo, try to just install the lib instead

azudindaem:
> @azudindaem that's because you've just cloned the repo, try to just install the lib instead

From there, you should begin. With a fresh start, without any interference from other libraries.

If there are any problems other than that, it is not from this library.

As a developer, you should know about the basic implementation.

```ts
import makeWASocket, { DisconnectReason } from '@adiwajshing/baileys'
import { Boom } from '@hapi/boom'

async function connectToWhatsApp () {
    const sock = makeWASocket({
        // can provide additional config here
        printQRInTerminal: true
    })
    sock.ev.on('connection.update', (update) => {
        const { connection, lastDisconnect } = update
        if(connection === 'close') {
            const shouldReconnect = (lastDisconnect.error as Boom)?.output?.statusCode !== DisconnectReason.loggedOut
            console.log('connection closed due to ', lastDisconnect.error, ', reconnecting ', shouldReconnect)
            // reconnect if not logged out
            if(shouldReconnect) {
                connectToWhatsApp()
            }
        } else if(connection === 'open') {
            console.log('opened connection')
        }
    })
    sock.ev.on('messages.upsert', m => {
        console.log(JSON.stringify(m, undefined, 2))

        console.log('replying to', m.messages[0].key.remoteJid)
        await sock.sendMessage(m.messages[0].key.remoteJid!, { text: 'Hello there!' })
    })
}
// run in main file
connectToWhatsApp()

```

You see nothing wrong with this statement. Can you read // can provide additional config here inside makeWASocket? We don't  need to put all the thing there. I don't want to argue with you about this matter.

PR #2677, opened 2023-03-31
fix: link preview

### LINKED ISSUE #2675: [BUG] Links Preview not working
Already reported in other issues:

#2603
#2390
[#2368 (Comment)](https://github.com/adiwajshing/Baileys/issues/2030#issuecomment-1220073936)
#1719
#1719

The issue still appears in the latest version of Baileys and wasn't fixed. 

This issue has the objective of documenting and keeping track of the problem.

FaizBastomi:
fix #2675
should now actually fix link preview not showing when sending text containing links with `http://` or `https://` prefix

### DIFF
--- a/src/Utils/link-preview.ts
+++ b/src/Utils/link-preview.ts
@@ -50,6 +50,7 @@ export const getUrlInfo = async(
 
 		const info = await getLinkPreview(previewLink, {
 			...opts.fetchOpts,
+			followRedirects: 'follow',
 			headers: opts.fetchOpts as {}
 		})
 		if(info && 'title' in info && info.title) {


azudindaem (commented):
Can you test with valid http or https but with invalid link?

FaizBastomi:
> Can you test with valid http or https but with invalid link?

@azudindaem 
is this what you mean? after trying, the link preview is missing because the link is invalid, or not accessible by my internet connection

<details open>
<summary><h3>Screenshots</h3></summary>
<p>

- the link preview is missing, because the link is invalid
image

- error due to invalid link
image

</p>
</details>

azudindaem:
Should be okay.. What is happen if we don't use followRedirects: 'follow'?

FaizBastomi:
> Should be okay.. What is happen if we don't use followRedirects: 'follow'?

@azudindaem 
sometimes when the link will be accessed with _link-preview-js_, they (website or link) redirect our request,
link-preview-js by default uses _error_ on `followRedirects`, so it fails many times when trying to generate link previews,
so, by default I force allow _link-preview-js_ to follow that redirect response, to minimize failure when trying to build link preview.

maybe this is not the best solution, if you have any suggestions, please let me know

src/Utils/link-preview.ts
```diff
@@ -50,6 +50,7 @@ export const getUrlInfo = async(
 
 		const info = await getLinkPreview(previewLink, {
 			...opts.fetchOpts,
+			followRedirects: 'follow',
```
adiwajshing, line 53: there should be a limit on the number of redirects followed, otherwise it can create a runaway loop

FaizBastomi: can you check again, I created a manual handle for redirection with limit of retries, if maximum retries is reached then process will return error

kamuridesu:
remove stale

kamuridesu:
remove stale

